PR #17, opened 2023-01-05
Problem: upgrading extensions that use shared memory or always-on workers

yrashk:
Every upgrade necessitates a Postgres server restart, which is not ideal.

Solution: introduce Dynpgext interface that abstracts these resources


yrashk:
It's not complete, but it is good enough to merge and continue iterating on to fill the gaps, test it more and add missing features.

PR #32, opened 2023-01-17
Problem: extensions are provisioned altogether

yrashk:
This pollutes the environment of extensions that potentially creates issues during tests or experimentation.

Solution: isolate every extension in its own share dir

As of now, this does not handle dependencies (REQUIRES) or ad-hoc combinations of extensions for manual testing.

The reason why omni_ext tests work is because we hardcoded this dependency in add_postgresql_extension. This whole thing needs to be improved.


yrashk:
Stale

PR #33, opened 2023-01-17
Problem: CI actions take some time (not too bad, but still some)

yrashk:
Solution: cache CPM dependency sources


yrashk:
CI build times are *okayish* for the time being.

PR #39, opened 2023-01-27
Problem: Omnigres adoption & licensing

yrashk:
If Omnigres will be growing in popularity, some of its functionality
may be considered for eventual inclusion into PostgreSQL.

However, our license, while fairly liberal, may be a concern because
it does not match the license of PostgeSQL.

Solution: provide a license exception for PostgreSQL

This way, if the code is to be included into PostgeSQL, it's licensed
under PostgeSQL license as well.


yrashk:
Side note: of course, as is, Omnigres code wouldn't be adoptable in Postgres (it is not C89), but it may be with just mechanical translation of it to C89 and applying Postgres formatting guidelines.

PR #55, opened 2023-02-09
Problem: omni_httpd's reload_configuration ends before the configuration is loaded

yrashk:
Solution: ensure the configuration is read before returning 
at the risk of setting the latch a number of times.


yrashk:
This isn't the right solution to the problem.

PR #72, opened 2023-03-03
Problem: docker images need to be built manually

proger:
Solution: propose an Action that makes builds and publishes to ghcr.io


yrashk:
Thank you! I need to figure out if this needs some configuration for repo/org to make ghcr.io packages work.

PR #75, opened 2023-03-03
Problem: using prepared Docker image on ARM CPUs

yrashk:
The prepared packaged is only built for linux/amd64

Solution: use cross-platorm capabilities to build for ARM64


yrashk:
It seems like the Docker build for arm64 hangs and gets eventually killed.

PR #81, opened 2023-03-06
Problem: docker layers caching in GHA

yrashk:
It takes too much space and the quota is rather limited (10GB)

Solution: cache the layers in S3


yrashk:
Superseded by #82

PR #94, opened 2023-03-10
Problem: writing cascading routing queries by hand

yrashk:
Writing something like this:

```sql
WITH test AS (SELECT omni_httpd.http_response(body := 'test') FROM request
              WHEREE request.path = '/test'),
     ping AS (SELECT omni_httpd.http_response(body := 'pong') FROM request WHERE request.path = '/ping')
SELECT * FROM test
UNION SELECT * FROM ping WHERE NOT EXISTS (SELECT FROM test)
```

is tedious and error-prone, especially when it comes to the union part.

Solution: make an aggregate function that simplifies this

### DIFF
--- a/extensions/omni_httpd/CMakeLists.txt
+++ b/extensions/omni_httpd/CMakeLists.txt
@@ -21,9 +21,9 @@ add_postgresql_extension(
         SCHEMA omni_httpd
         RELOCATABLE false
         SCRIPTS omni_httpd--0.1.sql
-        SOURCES omni_httpd.c master_worker.c http_worker.c fd.c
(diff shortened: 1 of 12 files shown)


yrashk:
There's still an issue with handling CTEs in supplied queries. If they are SELECTs, that's fine, they can stay inside of their respective queries, but if they are modifying statements, they need to go to the top level to work.

I am thinking to detect the modifying CTE queries and move them to the top. The only issue I have is name clashes. I can either leave this to the user to name them correctly or rename them upon conflict and fix the reference (would require deep node walking)

PR #100, opened 2023-03-12
Problem: omni_httpd config reload race condition

yrashk:
It's still possible to get a race condition. HTTP workers increment the semaphore upon initial statup. So when the master worker starts them, goes into the event loop and shortly after that reloads, it may see the correct (or even greater) number of HTTP workers on the semaphore.

Solution: HTTP workers should only increment semaphore on reload but not on the initial load.


yrashk:
Likely not a good fix. See #101 for a potentially better one.

PR #106, opened 2023-03-13
Problem: ocassionally role_name test fails

yrashk:
Instead of showing `test_user` it shows the name of the OS user (default username used for the database)

Solution: ensure we're not just checking pointers when comparing names

If the listener context is the same, then the check will never work correctly. For some reason, when I wrote this code, I thought that the pointers will never be the same.


yrashk:
I think #107 is better

yrashk:
#107 got merged

PR #110, opened 2023-03-14
Problem: having to update listener handle entry

### LINKED ISSUE #102: Problem: having to update listener handle entry
Any time one wants to modify request handling, they need to locate the correct handler for the listener and update the query in full.

However, this means that 

* the discoverability of handlers is obtuse (one needs to locate the listener and join with `handlers` over `listeners_handlers` to find the handler)
* the composability of such handlers is somewhat limited

Proposed idea #1: what if we introduced unique names to handlers so that they can be easily located?

Proposed idea #2: what if we added priority to `listeners_handlers` and took all joining listener handlers ordered by priority and combined them into one big query upon configuration reload?

yrashk:
Any time one wants to modify request handling, they need to locate the correct handler for the listener and update the query in full.

However, this means that

* the discoverability of handlers is obtuse (one needs to locate the listener and join with handlers over listeners_handlers to find the handler)
* the composability of such handlers is somewhat limited

Solution #1: introduce unique names to handlers so that they can be easily located

Solution #2: combine multiple listeners_handlers sorted by priority

Closes #102

---

TODO:

- [ ] Rewrite documentation to use this approach
- [ ] Figure out if `role_name` per handler and having handlers grouped by it is sensible

### DIFF
--- a/extensions/omni_httpd/CMakeLists.txt
+++ b/extensions/omni_httpd/CMakeLists.txt
@@ -23,7 +23,7 @@ add_postgresql_extension(
         SCRIPTS omni_httpd--0.1.sql
         SOURCES omni_httpd.c master_worker.c http_worker.c fd.c cascading_query.c
         TESTS_REQUIRE omni_ext
-        REGRESS http role_name cascading_query handler_validity port_selector)
+        REGRESS http cascading_query handler_validity port_selector)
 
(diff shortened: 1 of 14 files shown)


yrashk:
My current concern is that this forces a certain structure on HTTP handling and makes things like adding reusable CTEs problematic. Where do these go? Those aren't handlers.

Where as using `cascading_query` directly gives all the necessary freedom to how the query can be constructed. If one wants to augment it with reusable CTEs, they can use `omni_sql.add_cte`

PR #113, opened 2023-03-17
Problem: accessibility of Omnigres programming

yrashk:
While many people can use PL/pgSQL, not everybody can. It's not as popular as, say, JavaScript.

Solution: include plv8 into Omnigres container image

This gives users access to the V8 JavaScript engine and using it to interact with the database.

While at it, because I want to test it (having difficulty doing this locally) I've amended Docker CI workflow to publish suffixed branch images for PRs that come from our own repository.


yrashk:
Stale. Hoping to direct efforts toward our own runtime (JSC-based, perhaps)

PR #118, opened 2023-03-20
Problem: big queries in omni_httpd handler query

yrashk:
This is not a good practice to show.

Solution: extract such queries to stored procedures


yrashk:
Note: using stored procedures may not validate whether the stored procedure is actually going to work at runtime. Not sure what's the best course of action here, but at least for pl/pgSQL there's https://github.com/okbob/plpgsql_check/

PR #119, opened 2023-03-20
Problem: http_response is volatile

yrashk:
This may prevent some optimizations.

Solution: make it immutable

In my naive benchmarking, the performance gain was minimal (if any). Yet, perhaps, it's best to keep it immutable as it may still be better.


yrashk:
This is already so since bd18432ab0e2d99bf6459205684e066b5ae0e8a0

PR #132, opened 2023-03-24
Problem: can't lower omni_httpd handler role

yrashk:
It must be current_user and that's it. While one can change roles to set it, it is rather incovenient.

Solution: allow changing to roles that are accessible to the current_user


yrashk:
Not only this is inconvenient, this also may be impossible as the target role might not be able to modify `handlers`

PR #135, opened 2023-04-01
Problem: omni_httpd workers are blocking

### LINKED ISSUE #40: Problem: omni_httpd workers are blocking
Every HTTP worker is currently going busy when getting a new connection and a request. 

This works (and having more than one worker ensures there are more workers to accept new connections), however, it negatively impacts HTTP/2 multiplexing. Once the worker is processing the request, it will not be processing other requests.

Proposed solution: consider running the event loop in a separate thread (either always or when the worker is busy), proxying all other requests to the same address as the listening socket so that other workers can pick it up. There's a bit of overhead for this proxying, but perhaps it is better than just being able to process a single request at a time.

yrashk:
Every HTTP worker is currently going busy when getting a new connection and a request.

This works (and having more than one worker ensures there are more workers to accept new connections), however, it negatively impacts HTTP/2 multiplexing. Once the worker is processing the request, it will not be processing other requests.

Solution: run HTTP processing event loop in a separate thread and use the primary thread for maintenance operations and executing Postgres handlers.

In addition, if there are requests in flight (defined as Postgres executing something), and the protocol is 2 or above, proxy the request back. It's not perfect as at times it is faster to wait than reprocess the query elsewhere, but we'd need a more sophisticated mechanism to handle this. Or maybe more tuning knobs.

Resolves #40

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Install OpenSSL
+    - name: Install necessary dependencies
+      # OpenSSL and `timeout` (from coreutils)
       if: matrix.os == 'macos-12'
       env:
         HOMEBREW_NO_AUTO_UPDATE: "1"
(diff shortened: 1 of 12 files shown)


yrashk:
To note: this brings a bit of performance degradation as measured by, say, `wrk`. Approximately 20-25% slower. But it enables a very important scenario as described and paves the way to delayed HTTP responses, WebSockets, etc.

I am not giving up a hope that it can be sped up.

PR #145, opened 2023-04-10
Problem: sum types can't be toasted

yrashk:
They are declared as PLAIN and can't be TOASTed

Solution: declare variable-sized sum types EXTENDED and ensure to use detoasting when retrieving them.


yrashk:
#146 includes a followup improvement

PR #153, opened 2023-04-14
Problem: sequence counters do not get replicated

yrashk:
From Postgres documentation on [restrictions of logical replication](https://www.postgresql.org/docs/current/logical-replication-restrictions.html)

```
Sequence data is not replicated. The data in serial or identity columns
backed by sequences will of course be replicated as part of the table,
but the sequence itself would still show the start value on the
subscriber. If the subscriber is used as a read-only database,
then this should typically not be a problem. If, however, some
kind of switchover or failover to the subscriber database is
intended, then the sequences would need to be updated to the
latest values, either by copying the current data from the
publisher (perhaps using pg_dump) or by determining a
sufficiently high value from the tables themselves.
```

Solution: prefix local counter values with a node id / system identifier

This way each replica maintains its own counters but IDs are prefixed with the node's identifier.

This is somewhat less efficient as it stores and allocates a bit more data.

The implementation is not perfect and is not grealy optimized yet. At the very least, there's potential for packing smaller pairs that fit into `Datum` into passing by values instead of pointers.

As part of this commit, I've added Jinja library wrapper that is used to (manually) generate heaps of repetitive code. This is not automated yet as doing this with CMake was too much of a rabbit hole for the time being. In the initial implementation, I was doing all the initialization using C API (to accommodate the repetitivesness), but it was getting large and not very easy to follow, so I had to go back to defining it in C.


yrashk:
I have a concern about the `prefix_seq` structure. Depending on the sizes chosen, it may have padding, which means it'll waste some space when stored.

Options:

#### Let it waste space

Keep it as is and understand the wastage. You only incur it if you use unequally sized pairs.

#### Make it a packed structure

This may be slower to work with on architectures that allow this and an exceptional situation (crash) on others.

##### Measure performance differences and _cautiously_ allow packed structures on platforms that allow this? 

It gets trickier with the `internallength` setup from SQL. I would need to revert setup method from SQL back to C.

Maybe introduce a `prefix_seq_int16_int32_packed` type? Again, perhaps only on architectures that would not have a problem with non-aligned offsets.

#### Remove pairs that get padded

The pairs are:

16:32, 16:64, 32:16, 32:64, 64:16, 64:32

This leaves us with the following:

16:16, 32:32, 64:64

(equal sizes only)

##### Bonus: remove all but one pair (64:64?)

#### Fix the size to 128 bits

Effectively, always size for 64:64. It wastes space for anything below 64:64 but provides a _very theoretical_ opportunity for changing the type on the fly (though practically will likely require updating old rows)

yrashk:
I also don't _love_ `prefix_seq` naming. I don't like the name itself ("prefix_seq") and the lack of brevity in the naming schema: `prefix_seq_int16_int32`

For the schema, maybe other names can be used?

1. int16 -> small, int32 -> int, int64 -> big
2. int16 -> smallint, int32 -> int, int64 -> bigint
3. int16 -> s, int32 -> i, int64 -> d
4. 3. int16 -> s, int32 -> i, int64 -> b

1 and 2 are not much shorter or better. 3 and 4 are definitely cryptic.

`prefix_seq_sb` is shorter than `prefix_seq_int32_int64` but to somebody who doesn't know, it is indecipherable.

As for the name itself, other options:

1. `omni_seq.seq`
2. `pseq` (prefixed)
4. `dseq`, `distseq` or `dist_seq` (distributed)
6. `omni_seq` (omni-present sequence)
7. `shardseq` or `sharded_seq` (sharded sequence)

So far, `omni_seq` (or, fully-qualified `omni_seq.omni_seq`) is interesting because it kind of indicates that this sequence can be everywhere throughout the network cluster.

yrashk:
With regard to alignment/packing issue, I think this is what I am going to try:

1. Make the structure packed
2. Reorder prefix and val fields by their size. The biggest one goes first and is always self-aligned.
3. Static assertion will check that the second field (and, for uniformity, the first one) is self-aligned by checking that its offset can be divided by its alignment 
4. Static assertion will check that the struct is sized as a sum of its sizes.

The reordering and packing here should ensure that no trailing padding is done as we're not going to use stride addresses (we're not going to put these structures into an array)

yrashk:
Another concern here is that the syntax of making a column use omni sequence is verbose and requires a pre-created sequence.

It's not exactly a show-stopper, but if it can be made better, then it should be.

yrashk:
Perhaps the concept should not be called "<anything> sequences", because sequences are still the same (local sequences), but perhaps "<something> identifier"?

1. "prefixed identifier"
2. "distributed identifier"
3. "omni identifier"

PR #158, opened 2023-04-19
Problem: using role_name in omni_httpd.handlers

yrashk:
This refers to roles by name, which are resolved to OIDs during configuration reload.

This is potentially problematic as it means the role can be unavailable at the time of the configuration reload.

It can be fragile if something is to happen to the search path in the future.

Solution: make `role_name` become `role` of type `regrole`

It's not *substantially* better but it feels like it's more of the right thing to do.

This can enable us to use `pg_shdepend` to ensure we can't drop the roles that are used.

### DIFF
--- a/extensions/omni_httpd/docs/security.md
+++ b/extensions/omni_httpd/docs/security.md
@@ -1,3 +1,5 @@
+<!-- @formatter:off -->
+
 # Security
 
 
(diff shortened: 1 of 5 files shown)


yrashk:
If we still want this change, the tests need to be rewritten to the new format (pg_yregress).

yrashk:
No longer relevant

PR #168, opened 2023-04-24
Problem: multiple test cases in pg_yregress

yrashk:
Having to write a lot of cases or tons of data points to make a test is at times tedious.

Solution: implicitly support Inja templates to generate tests (or anything, really)

The trade-off in this approach is that you will not be able to use results of `pg_yregress` to overwrite the original file when you're updating tests. But that has always been a bit of an issue as it changed formatting and seemingly eliminated comments.

### DIFF
--- a/extensions/omni_seq/test.yml
+++ b/extensions/omni_seq/test.yml
@@ -120,22 +120,15 @@ tests:
 
 - name: smoke test
   steps:
-    - query: create sequence seq
-    - query: |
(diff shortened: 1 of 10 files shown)


yrashk:
I am not loving this because it breaks the YAML format and introduces a new language of a sort.

What if we used SQL to generate cases instead? It's right there.

yrashk:
Too stale.

Issue #177, opened 2023-05-11
Problem: omni_httpd will simply log an error if can't listen on a port


yrashk:
This means that one will have to look into the logs to see what happened and adjust accordingly.

Proposed solution: pick an unused port and record the effective port 
in `effective_port` like we do when using the wildcard (0) port.

idoqo:
Hi @yrashk since https://github.com/omnigres/omnigres/pull/191/ has been merged, is this still valid though?

PR #179, opened 2023-05-12
Problem: integrating with external services

yrashk:
There are some extensions out there to call HTTP but they have their own limitations, such as using cURL (which means we can't use a more efficient polling mechanism) or they way they handle async.

Solution: start developing omni_httpc based on libh2o

This doesn't feature the async client just yet as we're just starting, but it's coming.

I've also split omni_http out for shared functionality.

---

TODO

- [ ] ~~make sure role.yml test is not flaky / doesn't depend on `pg_sleep`~~(https://github.com/omnigres/omnigres/issues/186)


steve-chavez:
Hey @yrashk,

> cURL (which means we can't use a more efficient polling mechanism)

Could you elaborate on the above?

---

I heard about the `h2o` server before but not `libh2o`. It looks like a promising replacement for `libcurl`-based http clients.

I would be keen to contribute solving some low-hanging fruits to learn more about `libh2o` and Omnigres :slightly_smiling_face:

yrashk:
Hey @steve-chavez ,

Great seeing you here!

Generally, the original motivation for this choice is that there's libh2o there already. Also, to the best of my understanding, the default way to use curl multi is through poll(), which isn't necessarily the best approach. From what I understand, it is possible to integrate curl with epoll/kqueue, but libh2o already has those.

I also find some performance/latency reduction ideas in libh2o interesting.

As for this area of work, on this branch, I'm currently rewriting omni_httpd tests to use omni_httpc (along with fixing some bugs I found in omni_httpc already). I want to merge this branch as soon as possible to start a new one to tackle async callback API (which will enable calling back into Postgres when there's a result; I'm very excited about this one)

A potentially low hanging fruit on this thread (in the published code), is that it'll routinely time out when repeating requests to the same target. And it is also not reusing pools/connections when calling back again successfully (that would make it more performant). Does this interest you at all? I might be able to suggest other ideas if not.

I'm traveling this week, so my coding is relatively slow.

yrashk:
@steve-chavez Here's my current WIP on this branch's work: https://github.com/yrashk/omnigres/commits/omni_httpc-wip

Outside of the immediate area of omni_httpc but related to the whole HTTP stack, there are some other low(er?) hanging fruits. One can I immediately think of:

https://github.com/omnigres/omnigres/issues/177

steve-chavez:
> the default way to use curl multi is through poll(), which isn't necessarily the best approach. From what I understand, it is possible to integrate curl with epoll/kqueue, but libh2o already has those.

Oh yeah yeah. You can use libcurl's multi interface for integrating with epoll. TBH it's pain. And if you need an event loop you have to integrate something like libuv.

I've just noticed that libh2o has its own [event loop](https://powerdns.org/libh2o/#eventloops(libuv,native)). That's definitely a win over libcurl!

> And it is also not reusing pools/connections when calling back again successfully (that would make it more performant). Does this interest you at all?

Most definitely! But at this point with my current knowledge I might just slow you down. 

> Outside of the immediate area of omni_httpc but related to the whole HTTP stack, there are some other low(er?) hanging fruits. One can I immediately think of:
> https://github.com/omnigres/omnigres/issues/177

Cool. I'll give that one a shot at the end of the week. Thanks!

yrashk:
This took a while, but the basics seem to work now, and I am getting pretty happy with its capabilities, particularly the ability to make multiple requests in one go, like so:

```sql
select 
    version >> 8 as http_version, status, headers 
from
    omni_httpc.http_execute_with_options(omni_httpc.http_execute_options(http2_ratio => 100), 
        omni_httpc.http_request('https://example.com'),
        omni_httpc.http_request('https://example.org'))
```

It's not yet perfect, but it's getting close to the point where I am ok to merge it and iterate.

PR #192, opened 2023-06-03
Problem: trying to get non-existent header attribute

yrashk:
Before acc3973d34b012d7ba988fa355344a69a70f1707, there was an `append` attribute in headers, but it is gone now. However, it was still being retrieved. Valgrind pointed that out.

Solution: make it always set headers

I am not sure why `append` was removed. Unfortunately, I didn't record this decision in the aforementioned change.


yrashk:
Well, most likely, `append` was removed so that `http_header` can be extracted to `omni_http` to be used both by the client and the server. Question is: do we need to find a way to restore this functionality in the server?

PR #196, opened 2023-06-05
Problem: pg_yregress bundles steps into top-level tests

yrashk:
This makes reading TAP outputs more difficult.

Solution: use TAP14's subtests


yrashk:
Docker builds failing because of the rate limiting.

PR #209, opened 2023-06-16
Problem: extensions can't be built individually

yrashk:
This is because of their assumption that they are built in-tree. However, this would be very useful for packaging to avoid that.

Solution: fix build process to accommodate this need

Also, test that this works in CI.


yrashk:
The tests pass but there's a serious deficiency: it is breaking down if you change CMakeLists and re-run `cmake`, presumably because the guard flags used there preclude them from being re-included.

PR #217, opened 2023-07-04
Problem: macos build is broken on GitHub Actions because of homebrew

yrashk:
Solution: try upgrading to macos-13 runner first

(I've also reformatted that YAML a bit)


yrashk:
Superseded by #218

PR #230, opened 2023-08-01
Problem: successful tests don't fail if success: false

yrashk:
Solution: fix pg_yregress to set success to true
so that it fails if the expectation for the test is to fail.


yrashk:
So far, this is failing on Postgres below version 15 as `make_absolute_path` and `canonicalize_path` have changed significantly https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=c10f830c511f0ba3e6f4c9d99f444d39e30440c8

PR #260, opened 2023-09-13
Problem: report of failing to check httpd handler validity

yrashk:
This happens when handlers are updated. They are later reported by http worker as invalid.

Solution: ensure we're checking the new record when updated not the old one.

:facepalm:


yrashk:
cc @Dipenbhatt03 thank you for the original report!

PR #269, opened 2023-09-22
Migrate omni_sql tests to pg_yregress

idoqo:
This rewrites the existing `omni_sql` tests to use pg_yregress' format. It also removes the existing sql/* and *.out files from `omni_seq`, since those were already re-written in pg_yregress.


extensions/omni_seq/tests/test.yml
```diff
@@ -126,7 +126,7 @@ tests:
           omni_seq.id_int64_nextval(10, 'seq')
         )
     - query: |
-        insert into t select from generate_series(1, 10) returning id
+        insert into t select from generate_series(0, 10) returning id
```
idoqo, line 129: I noticed the original sql tests started the series from 0, I don't know if it's necessary to do that here so I can revert if it's not.

yrashk:
Michael, thank you so much for your contribution!

I amended it with two minor additions:

1. When we remove pg_regress tests, we must remove them from REGRESS lines in CMakeLists.txt; otherwise, `cmake` will fail. Essentially, if we remove all pg_regress tests, the REGRESS line must go https://github.com/omnigres/omnigres/pull/269/commits/5c97124447adcc0c9ce51d0fa2505080015818c4
2. CTE tests were together with other tests on the same level, making it harder to find their boundary. I've used [test grouping feature](https://docs.omnigres.org/pg_yregress/usage/#grouping-tests) in pg_yregress to reorganize: https://github.com/omnigres/omnigres/pull/269/commits/effedc4c74d748452634053aa20afd4510081b32

extensions/omni_seq/tests/test.yml
yrashk: I made this change to avoid oddly looking `11` there. I will revert this, but thanks for bringing attention to it! Details do matter.

yrashk:
I made a few more minor changes and merged them in. Thank you again and congratulations with your first contribution 🥳

PR #278, opened 2023-10-10
Problem: Multiple existing tests still uses pg_regress

idoqo:
Solution: Rewrite existing omni_types and omni_ext tests with pg_yregress. We also use yaml anchors for the SQL rewrites of psql commands such as \dT. There are still a few tests to migrate in libpgaug and omni_ext_tests (mostly ones involving more than one database).


yrashk:
Thank you for your contribution! Great start.

For the `\dT` command, you can start by poking around https://www.postgresql.org/docs/current/catalog-pg-type.html

yrashk:
Also, you mention tests for `omni_ext`, but it looks like they are not a part of this PR.

extensions/omni_types/tests/sum_type.yml
```diff
@@ -0,0 +1,652 @@
...
+    results:
+    - sum_type_recv: 'integer(1000)'
+
+  - name: Malformed input
+    query: savepoint try
+  - query: select sum_type_recv(E''::bytea);
+    error:
+      severity: ERROR
+      message: input is too short to fit a discriminant
+  - query: rollback to savepoint try
```
yrashk, line 465: we don't need savepoints anywhere here as it's all transactional anyway

extensions/omni_ext/tests/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+      
+#CREATE DATABASE another_db;
+#\connect another_db
+#CREATE EXTENSION omni_ext_test CASCADE;
+#NOTICE:  installing required extension "omni_ext"
+#SELECT omni_ext_test.alloc_shmem_database_local();
+# alloc_shmem_database_local 
+#----------------------------
+# testdb 1
+#(1 row)
```
idoqo, line 57: @yrashk Was wondering what would be the ideal way to test this - thought of using a different instance but I guess that defeats the test purposes (which is to create another db).

yrashk: Yeah, it's a good find! If we can't find a solution, I guess we can keep this test intact (that will keep the pg_regress test harness, which is not ideal)

Creating another instance here is beyond the point, as we're testing cross-database behavior in a single instance.

Currently, we can't do `create database` because we're wrapping everything into a transaction (`CREATE DATABASE cannot run inside a transaction block` error), and we can't switch databases on the fly (pg_yregress doesn't have support for it.

To address this, we can perhaps add the following functionality:

## Non-transactional tests in pg_yregress

Something like this?

```yaml
- query: create database another_db
   transaction: false
```

### Ability to send queries to other databases

This one is tougher to design for. Syntax-wise this can be as simple as:

```yaml
- query: select omni_ext_test.alloc_shmem_database_local()
   database: another_db
```

But we would need to connect to manage another connection.

---

If the above will take some time, it may be prudent to merge this PR (assuming it's translating old tests correctly) and open an issue tracking this commented-out test.

extensions/omni_ext/tests/tests.yml
```diff
@@ -0,0 +1,130 @@
...
+        
+
+  - name: database local bgworker
+    instance: with_omni_ext_test
+    steps:
+    - query: select omni_ext_test.wait_for_table('local_worker_started')
+      results:
+      - wait_for_table: true
+    
+    #\d global_worker_started;
```
yrashk, line 87: is this (and below) a placeholder for something or just a comment?

yrashk:
Also, it looks like libpgaug/test still relies on pg_regress (if the intention was go get rid of all pg_regress tests in this PR)

extensions/omni_ext/tests/tests.yml
idoqo: Right - wanted to get your thoughts on it too. We are describe the table, but not expecting a value. We can re-implement `\d` in SQL - but it could take some effort, so I wanted to know if it was a really necessary test then I can go ahead and do that.

steve-chavez: > We can re-implement \d in SQL - but it could take some effort

Just in case it helps. You can do `\set ECHO_HIDDEN` in `psql` to get the queries executed by any slash command (like `\d`).

extensions/omni_ext/tests/tests.yml
yrashk: Extracted this to a bounty: https://github.com/omnigres/omnigres/issues/351

yrashk:
Interesting CI failure on macOS:

```yaml
expected:
    query: select row ('John', '01/10/1980')::person::sum_type
    results:
     - row: person((John,1980-01-10))
result:
    query: select row ('John', '01/10/1980')::person::sum_type
    results:
    - row: "person((John,1980-10-01))"  
 ```

The difference is in how the date is being processed. This test is not failing on my local macOS. While we can fix the environment on the CI macOS, that may be too fragile. Perhaps it is better to rewrite this test to be less fragile when parsing that date?

idoqo:
Yeah, saw the CI failure but was unsure since it was all green on Linux and we've had issues with macOS CI in the past.

> Perhaps it is better to rewrite this test to be less fragile when parsing that date?

Agreed, I've updated it to use `to_date` for parsing in the query, hopefully that makes it more deterministic?

yrashk:
It has resolved the CI issue indeed!

Issue #279, opened 2023-10-12
Problem: some HTTP servers require Content-Length to be set to 0


yrashk:
Currently, omni_httpd only sets `Content-Length` if the body is larger than zero bytes:

https://github.com/omnigres/omnigres/blob/master/extensions/omni_httpc/omni_httpc.c#L482

This is done similarly to H2O's HTTP client:

https://github.com/h2o/h2o/blob/master/src/httpclient.c#L484-L489

The HTTP/1.1 spec says:

<blockquote>
Applications SHOULD use this field to indicate the transfer-length of
   the message-body, unless this is prohibited by the rules in [section](https://www.rfc-editor.org/rfc/rfc2616#section-4.4)
   [4.4](https://www.rfc-editor.org/rfc/rfc2616#section-4.4).

...

   For compatibility with HTTP/1.0 applications, HTTP/1.1 requests
   containing a message-body MUST include a valid Content-Length header
   field unless the server is known to be HTTP/1.1 compliant. If a
   request contains a message-body and a Content-Length is not given,
   the server SHOULD respond with 400 (bad request) if it cannot
   determine the length of the message, or with 411 (length required) if
   it wishes to insist on receiving a valid Content-Length.

</blockquote>

Proposed solution 1: always include `Content-Length` (unless we're doing chunked transfers, of course)

Proposed solution 2: same as above, but provide an optional flag to suppress `Content-Length: 0` in the name of compaction

steve-chavez:
> Currently, omni_httpd only sets Content-Length if the body is larger than zero bytes:

Hm, the above referred to `omni_httpc` and not `omni_httpd` right? I've tested `omni_httpd` and it always sets the `Content-Length` despite an empty response body. This is done here:

https://github.com/omnigres/omnigres/blob/e5ede5f0936ec1e02d765e557a405bfbc3a1fda3/extensions/omni_httpd/http_worker.c#L774

Would a test ensuring this behavior (`content-length` always set) be useful?

yrashk:
You are right, the issue is about `omni_httpc`! That was my mistake, and thanks for spotting it!

yrashk:
> Would a test ensuring this behavior (`content-length` always set) be useful?

On the httpc or the httpd side? Or both?

steve-chavez:
>  httpd side?

Was thinking one for the `httpd` side first. Just to get me started :)

yrashk:
I think there may be some interesting corner cases to test and codify.

[...]

PR #314, opened 2023-10-29
Problem: omni_httpd content-length behavior is not clear

steve-chavez:
omni_httpd currently sets it automatically for any non-null response body.

Solution: document it and add a test for null response body behavior.

----

Related to the discussion on https://github.com/omnigres/omnigres/issues/279#issuecomment-1784135868

### DIFF
--- a/extensions/omni_httpd/docs/headers.md
+++ b/extensions/omni_httpd/docs/headers.md
@@ -45,34 +45,10 @@ Content-Type: text/plain
 Hi there
 ```
 
-If you set a lower `Content-Length`, the response body will be resized accordingly:
+!!! warning "Overriding Content-Length"
 
-```sql
-update
-omni_httpd.handlers
-set query = $$
-  select *
-  from omni_httpd.http_response(
-    headers => array [omni_http.http_header('Content-Type', 'text/plain'), omni_http.http_header('Content-Length', '2')],
-    body => 'Hi there'
-  )
-$$
-where id = 1;
-```
+    `omni_httpd` allows overriding the `Content-Length`. This is useful for integrating with other HTTP handlers (e.g. Flask) that set the `Content-Length`.
+    To ensure correctness, overriding works in the following way:
 
-```bash
-curl localhost:8080 -i
-
-HTTP/1.1 200 OK
-Connection: keep-alive
-Content-Length: 2
-Server: omni_httpd-0.1
-Content-Type: text/plain
-
-Hi
-```
-
-!!! warning "Overflowing Content-Length"
-
-    If you set a Content-Length higher than the size of the response body, `omni_httpd` will
-    emit a WARNING and defer to using the actual body size.
+    - If the `Content-Length` is set lower than the actual body size. `omni_httpd` will use the new `Content-Length` and downsize the response body.
+    - If the `Content-Length` is set higher than the actual body size. `omni_httpd` will keep its `Content-Length`, emit a WARNING and use the actual body size.
--- a/extensions/omni_httpd/docs/headers.md
+++ b/extensions/omni_httpd/docs/headers.md
@@ -9,3 +9,72 @@ select omni_http.http_header_get_all(request.headers, 'accept') as accept;
 ```
 
 The header name these functions take is case-insensitive.
+
+## Content-Length
+
+`omni_httpd` automatically sets the `Content-Length` header for any non-null response body.
+
+For instance, having:
+
+```sql
+create extension omni_httpd cascade;
+
+set omni_httpd.init_port to 8000;
+
+update
+omni_httpd.handlers
+set query = $$
+  select *
+  from omni_httpd.http_response(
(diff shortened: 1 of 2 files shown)


yrashk:
Thank you so much for your contribution! 🤟 

Resuming our conversation in #279:

1. Do we want Content-Length to be absent if the body is `null`? 
2. As for the overflow, I am trying to think if this is more of a warning or error. I am inclined to believe this is a warning as we don't do anything crazy there, but it is indicative of a potential problem one wouldn't want to bury. I believe `omni_alerts` (when it comes to live) is the best way to alert ops of the problem, but whether that's a 500 or not, it is hard to say. Thoughts?

extensions/omni_httpd/docs/headers.md
```diff
@@ -8,4 +8,73 @@ select omni_http.http_header_get(request.headers, 'host') as host;
...
+## Content-Length
+
+`omni_httpd` automatically sets the `Content-Length` header for any non-null response body.
+
+For instance, having:
+
+```sql
+create extension omni_httpd cascade;
+
+set omni_httpd.init_port to 8000;
```
yrashk, line 22: This is ineffectual because only the extension installation script uses these parameters.

This can lead to confusion. Perhaps for documenting purpose, we can remove this (instead of moving it above `create extension`)

extensions/omni_httpd/docs/headers.md
```diff
@@ -8,4 +8,73 @@ select omni_http.http_header_get(request.headers, 'host') as host;
 select omni_http.http_header_get_all(request.headers, 'accept') as accept;
 ```
 
-The header name these functions take is case-insensitive.
\ No newline at end of file
+The header name these functions take is case-insensitive.
+
+## Content-Length
```
yrashk, line 13: Great idea to document this!

extensions/omni_httpd/docs/headers.md
steve-chavez: Removed. Now using the default `8080` port for the curl examples instead.

steve-chavez:
> Do we want Content-Length to be absent if the body is null?

It seems a bit wasteful to always include a `Content-Length: 0` when the body is known to be null. Also, the [RFC](https://www.rfc-editor.org/rfc/rfc9110#name-content-length) doesn't mandate setting `Content-Length` for empty bodies. So any behavior would be correct, but the `null` to no `Content-Length` might be more intuitive.

> As for the overflow, I am trying to think if this is more of a warning or error. I am inclined to believe this is a warning as we don't do anything crazy there, but it is indicative of a potential problem one wouldn't want to bury. I believe omni_alerts (when it comes to life) is the best way to alert ops of the problem, but whether that's a 500 or not, it is hard to say. Thoughts?

Taking a step back, why is `Content-Length` overridable? What use case would that cover?

`omni_httpd` could always compute the correct `Content-Length` and if a user sets it, we could emit a WARNING saying "ignoring user defined content-length". To resize the response, users could manipulate the body directly.

yrashk:
> > Do we want Content-Length to be absent if the body is null?
> 
> It seems a bit wasteful to always include a `Content-Length: 0` when the body is known to be null. Also, the [RFC](https://www.rfc-editor.org/rfc/rfc9110#name-content-length) doesn't mandate setting `Content-Length` for empty bodies. So any behavior would be correct, but the `null` to no `Content-Length` might be more intuitive.

Fair. I think we can agree on this.

> 
> > As for the overflow, I am trying to think if this is more of a warning or error. I am inclined to believe this is a warning as we don't do anything crazy there, but it is indicative of a potential problem one wouldn't want to bury. I believe omni_alerts (when it comes to life) is the best way to alert ops of the problem, but whether that's a 500 or not, it is hard to say. Thoughts?
> 
> Taking a step back, why is `Content-Length` overridable? What use case would that cover?
> 
> `omni_httpd` could always compute the correct `Content-Length` and if a user sets it, we could emit a WARNING saying "ignoring user defined content-length". To resize the response, users could manipulate the body directly.

This behavior was introduced very recently. Initially, I was handling a case when we had a case of HTTP handler producing its own content-length, and omni_httpd always set its own: 783ae3ade2bd99e96bf82f7b32421350dc29a9e4 and as a quick follow up, I wrote a test for it: e60deed42781c2e76ff393819fd72f25a7becfc3

The origin of this issue lies in frameworks such as, say, Flask, setting their own `content-length`. The reasoning I had was to basically respect what these frameworks say is the expected length and just as a precaution, resize the output when there's too much of it. May be that should also include a warning.

steve-chavez:
> The origin of this issue lies in frameworks such as, say, Flask, setting their own content-length. The reasoning I had was to basically respect what these frameworks say is the expected length and just as a precaution, resize the output when there's too much of it. May be that should also include a warning.

I see, makes sense. It looks like this behavior is not something that the user might use directly through SQL. So I've removed the downsizing example and left overriding documented under a warning. Let me know what you think.

yrashk:
I think it's perfect, at least for now! Thank you.

PR #316, opened 2023-10-30
Problem: Flask POST methods not passing headers & body

inevolin:
Flask POST handlers are missing headers and body data.

https://discord.com/channels/1059495229243265096/1168330166842425374

### DIFF
--- a/languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
+++ b/languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
@@ -5,6 +5,7 @@ from typing import TypedDict, Optional, Any
 from dataclasses import dataclass
 
 import sys
+import io
 
 @dataclass
 class HTTPRequest:
@@ -15,11 +16,11 @@ class HTTPRequest:
     headers: list[HttpHeader]
 
     def wsgi_environ(self) -> dict:
-        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': self.body or b"",
-                   'wsgi.multithread': False,
+        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""),
+                   'wsgi.input_terminated': 1, 'wsgi.multithread': False,
                    'REQUEST_METHOD': self.method, 'PATH_INFO': self.path, 'QUERY_STRING': self.query_string or ''}
         for header in self.headers:
-            environ[f"HTTP{header['name'].upper()}"] = header['value']
+            environ[f"HTTP_{header['name'].upper().replace('-', '_')}"] = header['value']
         return environ
 
     @classmethod
--- a/languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
+++ b/languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
@@ -16,11 +16,14 @@ class HTTPRequest:
     headers: list[HttpHeader]
 
     def wsgi_environ(self) -> dict:
-        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""),
-                   'wsgi.input_terminated': 1, 'wsgi.multithread': False,
+        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""), 'wsgi.multithread': False,
                    'REQUEST_METHOD': self.method, 'PATH_INFO': self.path, 'QUERY_STRING': self.query_string or ''}
         for header in self.headers:
             environ[f"HTTP_{header['name'].upper().replace('-', '_')}"] = header['value']
+        
+        # set content-length header to enable body stream
+        if not environ.get('CONTENT_LENGTH'): # TODO: how to handle chunked data?
+            environ['CONTENT_LENGTH'] = str(len(self.body))
         return environ
 
     @classmethod
--- a/languages/python/omni_http/src/omni_http/omni_httpd/flask.py
+++ b/languages/python/omni_http/src/omni_http/omni_httpd/flask.py
@@ -1,6 +1,6 @@
 from typing import Optional
 
-from . import HTTPResponse, HTTPRequest
+from . import HTTPResponse, HTTPRequest, HTTPOutcome
 from ..omni_http import HttpHeader
 
 try:
@@ -27,7 +27,7 @@ class Adapter:
     def __init__(self, app: flask.Flask):
         self.app = app
 
-    def __call__(self, req: HTTPRequest) -> HTTPResponse:
+    def __call__(self, req: HTTPRequest) -> HTTPOutcome:
         res = responder()
         response = self.app.wsgi_app(req.wsgi_environ(), res)
 
@@ -35,4 +35,4 @@ class Adapter:
         for chunk in response:
             body += chunk
 
-        return HTTPResponse(status = res.status, body = body, headers = res.headers)
+        return HTTPResponse(status = res.status, body = body, headers = res.headers).outcome()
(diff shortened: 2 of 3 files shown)


yrashk:
# Suggestion for this PR

How about we turn this PR into an opportunity to have a test suite for how Python/WSGI integration works? This can be very useful in ensuring this keeps working without breaking.

The interesting part is that we don't even necessarily need to run the HTTP server and a client in the test as we're just testing the conversions. We can simply use `omni_httpd.http_request` to form requests and throw them against Python-based handler functions. A lot of Python tests are done in omni_python tests.

Feel free to ping me on Discord if any help or further explanation would be useful.

### Minor note on the commit P/S messages

The general pattern I would suggest following (not a strict requirement per se, but I see a lot of value in it) is focusing on the actual problem first (not the solution described through a problem) and then the solution.

I wrote about this before: https://yrashk.com/blog/2017/09/04/solving-problems-one-commit-at-a-time/

README.md
```diff
@@ -127,8 +127,9 @@ Below is the current list of components being worked on, experimented with and d
 To build and run Omnigres, you would currently need a recent C compiler, OpenSSL and cmake:
 
 ```shell
-cmake -S . -B build
-cmake --build build --parallel
+cmake -DCMAKE_BUILD_TYPE=Release -S . -B build # -DCMAKE_BUILD_TYPE=Release
```
yrashk, line 130: This probably belongs to a separate PR.

languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
```diff
@@ -15,11 +16,11 @@ class HTTPRequest:
     headers: list[HttpHeader]
 
     def wsgi_environ(self) -> dict:
-        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': self.body or b"",
-                   'wsgi.multithread': False,
+        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""),
+                   'wsgi.input_terminated': 1, 'wsgi.multithread': False,
```
yrashk, line 20: It looks like https://peps.python.org/pep-3333/ doesn't define input_terminated. I wonder what the authoritative reference for this is?

inevolin: It is defined in a flask extension: https://tedboy.github.io/flask/_modules/werkzeug/wsgi.html

Alternatively I think we can also set the content-length header, but either way the logic for retrieving the stream is defined and used from this extension.

README.md
inevolin: removed for now

languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
yrashk: Interesting. Trying to find some more information if this is Flask-specific and if anything else uses this.

languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
```diff
@@ -15,11 +16,14 @@ class HTTPRequest:
...
-        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': self.body or b"",
-                   'wsgi.multithread': False,
+        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""), 'wsgi.multithread': False,
                    'REQUEST_METHOD': self.method, 'PATH_INFO': self.path, 'QUERY_STRING': self.query_string or ''}
         for header in self.headers:
-            environ[f"HTTP{header['name'].upper()}"] = header['value']
+            environ[f"HTTP_{header['name'].upper().replace('-', '_')}"] = header['value']
+        
+        # set content-length header to enable body stream
+        if not environ.get('CONTENT_LENGTH'): # TODO: how to handle chunked data?
```
inevolin, line 25: I'm leaving out Transfer-Encoding header, because I was not able to get that header. Every curl request with this header is just not being passed/captured for some reason. Maybe because the actual data was not chunked, so curl removes the header. This can be fixed in a future PR.

languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
inevolin: this has been changed to using `content-length` instead of `input_terminated`

languages/python/omni_http/src/omni_http/omni_httpd/__init__.py
```diff
@@ -15,11 +16,14 @@ class HTTPRequest:
     headers: list[HttpHeader]
 
     def wsgi_environ(self) -> dict:
-        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': self.body or b"",
-                   'wsgi.multithread': False,
+        environ = {'wsgi.version': (1, 0), 'wsgi.url_scheme': 'http', 'wsgi.input': io.BytesIO(self.body or b""), 'wsgi.multithread': False,
```
inevolin, line 19: the input stream is important, using bytes doesn't work

languages/python/omni_http/src/omni_http/omni_httpd/flask.py
```diff
@@ -27,12 +27,12 @@ class Adapter:
...
         res = responder()
         response = self.app.wsgi_app(req.wsgi_environ(), res)
 
         body = b""
         for chunk in response:
             body += chunk
 
-        return HTTPResponse(status = res.status, body = body, headers = res.headers).outcome()
\ No newline at end of file
+        return HTTPResponse(status = res.status, body = body, headers = res.headers)
```
inevolin, line 38: the python test script's `handle` function needs to be adjusted to:
`
@pg
def handle(req: omni_httpd.HTTPRequest) -> omni_httpd.HTTPOutcome:
    return app_(req).outcome()
`

inevolin:
@yrashk interesting, locally all tests succeed, but fail in this pipeline because of `omni_http[Flask]`

yrashk:
> @yrashk interesting, locally all tests succeed, but fail in this pipeline because of `omni_http[Flask]`

I think this one is easy:

```diff
diff --git a/extensions/omni_python/tests/omni_python_flask.yml b/extensions/omni_python/tests/omni_python_flask.yml
index 0ab1295ce..386f1891a 100644
--- a/extensions/omni_python/tests/omni_python_flask.yml
+++ b/extensions/omni_python/tests/omni_python_flask.yml
@@ -11,8 +11,8 @@ instance:
   - query: select omni_python.install_requirements($1)
     params:
     - |
-      omni_http
-      omni_python[Flask]
+      omni_http[Flask]
+      omni_python
 
 tests:
 
```

Chances are it worked locally because you may have head `flask` installed globally in your Python installation.

languages/python/omni_http/src/omni_http/omni_httpd/flask.py
yrashk: I think we can actually look into amending `omni_httpd` extension behavior.Right now it's looking for `http_outcome` only, but we can also let it look for `http_response` individually (if other responses are impossible; and may be allowed singling out other types of outcomes). This way no conversion would be necessary.

languages/python/omni_http/src/omni_http/omni_httpd/flask.py
```diff
@@ -27,12 +27,12 @@ class Adapter:
...
         res = responder()
         response = self.app.wsgi_app(req.wsgi_environ(), res)
 
         body = b""
         for chunk in response:
             body += chunk
 
-        return HTTPResponse(status = res.status, body = body, headers = res.headers)
\ No newline at end of file
+        return HTTPResponse(status = res.status, body = body, headers = res.headers).outcome()
```
yrashk, line 38: Should this be `HTTPResponse` for the testability we discussed?

inevolin: The test is still there in a way. This was reverted given your suggestion here - https://github.com/omnigres/omnigres/pull/316#discussion_r1376973487

yrashk: I may not have been 100% clear there. I meant perhaps we should look into making omni_httpd server accept http_response in addition to http_outcome. This way, Python stuff can return http response without any "outcomes" if it doesn't need to do things like reproxying.

Issue #320, opened 2023-11-02
Problem: omni_python is undocumented


yrashk:
This is a severe roadblock in its adoption.

Proposed solution: write usage documentation outlining the workflow

It should have a good amount of good, meaningful examples and feature Flask integration.

You can gather how to use it in these places:

* https://www.loom.com/share/b4701cbc77e941919cd5dbff7b15465e
* https://github.com/omnigres/omnigres/tree/master/extensions/omni_python/tests
* https://github.com/omnigres/omnigres/blob/master/extensions/omni_schema/tests/omni_python.yml
* [Discord](https://discord.omnigr.es)

---

**Bounty size: $250**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

Akshat-Jain:
@yrashk I'd like to claim this one. Can you assign it to me?

PR #328, opened 2023-11-04
Problem: omni_python is undocumented

Akshat-Jain:
Solution: Add examples demonstrating the usage of `omni_python`

Closes https://github.com/omnigres/omnigres/issues/320

### DIFF
--- a/extensions/omni_python/docs/intro.md
+++ b/extensions/omni_python/docs/intro.md
@@ -163,6 +163,12 @@ import uuid
 
 app = Flask('myapp')
 
+def result_to_json(result):
+    rows = []
+    for row in result:
+        rows.append(dict(row))
+    return json.dumps(rows)
+
 @app.route('/employees', methods=['POST'])
 def create_employee():
     from flask import make_response, request
@@ -177,17 +183,17 @@ def create_employee():
         return "Missing required fields", 400
 
     employee = plpy.execute(plpy.prepare("INSERT INTO employees (id, name, department, salary) VALUES ($1, $2, $3, $4) RETURNING *", ["uuid", "text", "text", "int"]), [employee_id, employee_name, employee_department, employee_salary])
-    return str(employee)
+    return result_to_json(employee)
 
 @app.route('/employees', methods=['GET'])
 def get_employees():
     employees = plpy.execute(plpy.prepare("SELECT * FROM employees"))
-    return str(employees)
+    return result_to_json(employees)
 
 @app.route('/employees/<employee_id>', methods=['GET'])
 def get_employee(employee_id):
     employee = plpy.execute(plpy.prepare("SELECT * FROM employees WHERE id = $1", ["uuid"]), [employee_id])
-    return str(employee)
+    return result_to_json(employee)
 
 app_ = flask.Adapter(app)
 
@@ -230,24 +236,37 @@ You can hit the endpoints defined in the Flask code above.
 Fetch all employees:
 ```shell
 $ curl http://localhost:5000/employees
-<PLyResult status=5 nrows=2 rows=[
-{'id': '3f59ef0a-1f57-405a-87d4-662f16698a72', 'name': 'Akshat', 'department': 'Engineering', 'salary': 200000}, 
-{'id': '77e52508-829e-4004-9ed5-7e386bb18d76', 'name': 'Daniel', 'department': 'Engineering', 'salary': 100000}
-]>
+[
+    {
+        "id": "7035dfea-6069-4426-bc16-87e1b90b3fc6",
+        "name": "Akshat",
+        "department": "Engineering",
+        "salary": 100000
+    },
+    {
+        "id": "5f23c592-0408-4ab1-bd19-ceedf3342b91",
+        "name": "Mohit",
+        "department": "Sales",
+        "salary": 50000
+    }
+]
 ```
 
 Create a new employee:
 ```shell
-$ curl -X POST http://localhost:5000/employees
-<PLyResult status=11 nrows=1 rows=[
-{'id': '8b26fcf2-4ba8-4b4a-8e9c-1bb08d0695e4', 'name': 'John Doe', 'department': 'Engineering', 'salary': 100000}
-]>
+$ curl --request POST \
+  --url http://localhost:5000/employees \
+  --header 'Content-Type: application/json' \
+  --data '{
+  "name": "Daniel",
+  "department": "Marketing",
+  "salary": 70000
+}'
+[{"id": "c919d532-fe9e-46ac-86e0-5f7d950fc1ea", "name": "Daniel", "department": "Marketing", "salary": 70000}]
 ```
 
 Fetch a particular employee:
 ```shell
-$ curl http://localhost:5000/employees/8b26fcf2-4ba8-4b4a-8e9c-1bb08d0695e4
-<PLyResult status=5 nrows=1 rows=[
-{'id': '8b26fcf2-4ba8-4b4a-8e9c-1bb08d0695e4', 'name': 'John Doe', 'department': 'Engineering', 'salary': 100000}
-]>
+$ curl http://localhost:5000/employees/c919d532-fe9e-46ac-86e0-5f7d950fc1ea
+[{"id": "c919d532-fe9e-46ac-86e0-5f7d950fc1ea", "name": "Daniel", "department": "Marketing", "salary": 70000}]>
 ```
--- a/extensions/omni_python/docs/intro.md
+++ b/extensions/omni_python/docs/intro.md
@@ -146,7 +146,7 @@ from
 Let's say you have a table called `employees`.
 ```postgresql
 create table employees (
-   id uuid,
+   id integer primary key generated always as identity,
    name text not null,
    department text not null,
    salary integer not null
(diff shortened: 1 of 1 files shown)


extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,196 @@
...
+def subtract(a: int, b: int) -> int:
+    return a - b
+```
+
+4. Create a function to define an identifier for the local directory you want to load into the database. For the example, the local directory is `python-files`.
+
+```postgresql
+create or replace function demo_function() returns omni_vfs.local_fs language sql
+as $$
+select omni_vfs.local_fs('/python-files')
```
yrashk, line 60: Just a note: the upcoming CLI tooling will take care of directory mapping.

Not sure we need to do anything about this here, but I was thinking maybe we can leave a tip there (there`s tip admonition syntax in mkdocs: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:tip). Thoughts on whether that's useful at all?

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,196 @@
...
+```python
+from omni_python import pg
+from omni_http import omni_httpd
+from omni_http.omni_httpd import flask
+from flask import Flask, jsonify, make_response
+
+app = Flask('myapp')
+
+@app.route("/")
+def ping():
```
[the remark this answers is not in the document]
akjn-yb: I read in some past thread that we don't have the DB API ready for Flask right now?

I was initially thinking of having the Flask app store some DB record and fetch it, any hints on that?

Or any other suggestions on what kind of examples you'd like to have?

I felt like the examples you demonstrated showcase the basic functionalities quite nicely, so re-used those.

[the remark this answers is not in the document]
yrashk: This is correct that we do not have DB API support just yet (this is completely orthogonal to Flask integration) and of course, we want that ASAP.

But, in the short term, we can still use [plpython's own API for database access](https://www.postgresql.org/docs/current/plpython-database.html) and perhaps leave a comment there saying that this is to be replaced by DB API compatible APIs and/or other frameworks (such as SQLAlchemy) when this will be ready, to give people an idea of where this is heading.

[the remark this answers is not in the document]
yrashk: As for examples, to me they feel messy because they were literally meaningless "somethings" I quickly put together.  If we can create a super small mini app that does something useful (write something to the database and read back in another point, or something to that end), that'd be fantastic.

To make sure it really works we need to get #316 in as it fixes rather important issues.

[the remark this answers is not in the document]
Akshat-Jain: @yrashk Have added a better example, and left todo comment where we had a dependency on the open PR. Can take it up separately later.
Hope that works!

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,196 @@
...
+def show_post(post_id):
+    resp = make_response(f'<h1>Post #{post_id}</h1>')
+    resp.set_cookie('cookie', 'yum')
+    return resp
+
+@app.route("/test.json")
+def json():
+    return jsonify({"test": "passed"})
+
+app_ = flask.Adapter(app)
```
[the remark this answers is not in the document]
Akshat-Jain: Sure, have added some details. Please have a look, and let me know if anything needs correction. Thanks!

extensions/omni_python/docs/intro.md
Akshat-Jain: Have added a tip conveying this. Hope that works!

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,196 @@
...
+```postgresql
+create or replace function demo_function() returns omni_vfs.local_fs language sql
+as $$
+select omni_vfs.local_fs('/python-files')
+$$;
+```
+
+5. Configure `omni_python`
+```postgresql
+insert into omni_python.config (name, value) values ('pip_find_links', '/python-wheels');
```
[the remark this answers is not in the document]
Akshat-Jain: Same as the other comment. Have added a tip conveying this.

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,244 @@
...
+from omni_http.omni_httpd import flask
+from flask import Flask, jsonify, make_response
+import uuid
+
+app = Flask('myapp')
+
+@app.route('/employees', methods=['POST'])
+def create_employee():
+    employee_id = uuid.uuid4()
+    # todo: Use actual data from request body once https://github.com/omnigres/omnigres/pull/316 is merged.
```
yrashk, line 169: #316 is in

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,272 @@
...
+    handler.id
+from
+    listener,
+    handler;
+```
+
+Let's say you have a table called `employees`.
+```postgresql
+create table employees (
+   id uuid,
```
yrashk, line 149: Minor nitpicking:

I/we generally don't recommend `uuid` for the `id` columns.

Throught out the project we use

```sql
 id                       integer primary key generated always as identity,
```

For user applications, it's often recommended to bump `int` to `int8` for practical purposes:

```sql
 id8                       integer primary key generated always as identity,
```

The idea here is to teach good practices consistently.

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,272 @@
...
+
+    employee_id = uuid.uuid4()
+    employee_name = json_data.get('name')
+    employee_department = json_data.get('department')
+    employee_salary = json_data.get('salary')
+
+    if not employee_name or not employee_department or not employee_salary:
+        return "Missing required fields", 400
+
+    employee = plpy.execute(plpy.prepare("INSERT INTO employees (id, name, department, salary) VALUES ($1, $2, $3, $4) RETURNING *", ["uuid", "text", "text", "int"]), [employee_id, employee_name, employee_department, employee_salary])
```
yrashk, line 185: Minor nitpicking again:

1. Consider breaking up the long line into multiple lines for readability
2. Our SQL style is generally lower-case (that said, we didn't publish a style guide yet :)
3. This should probably be called `employees` not `employee`

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,272 @@
...
+    return result_to_json(employee)
```
yrashk, line 186: Nitpicking: I haven't checked my code, but it may be more elegant/Pythonesque to write something like 

```python
return json.dumps([dict(employee) for employee in employees])
```

Whether here or in `result_to_json` (`employees_to_json`?)

extensions/omni_python/docs/intro.md
```diff
@@ -0,0 +1,272 @@
...
+    for row in result:
+        rows.append(dict(row))
+    return json.dumps(rows)
+
+@app.route('/employees', methods=['POST'])
+def create_employee():
+    from flask import make_response, request
+    json_data = json.loads(request.data.decode('UTF-8'))
+
+    employee_id = uuid.uuid4()
```
yrashk, line 177: if the `uuid` is replaced with a traditional `int8` that is generated, no need in this.

extensions/omni_python/docs/intro.md
Akshat-Jain: > This should probably be called employees not employee

Which one? If you mean the variables like `employee_id`, I think `employee` seems fine as these pertain to a single employee. Please let me know what I'm missing, thanks!

yrashk: I think I got confused because it was then proceeding to list processing with `result_to_json` (or comprehension as suggested). Becase it does technically return a list. What if we did `employee = ...[0]` and then convert to json just for that one dict? What looks/feels better to you?

extensions/omni_python/docs/intro.md
Akshat-Jain: > I/we generally don't recommend uuid for the id columns.

Just curious about this: Why?

extensions/omni_python/docs/intro.md
Akshat-Jain: > What if we did employee = ...[0] and then convert to json just for that one dict? 

That would either require in-line logic, or another function `employee_to_json` IIUC? If yes, seems unnecessary to me.

The current `employees_to_json()` is usable in all the endpoint responses, which looks cleaner to me.

extensions/omni_python/docs/intro.md
yrashk: Along these lines:

* https://medium.com/@sjksingh/postgresql-primary-key-dilemma-uuid-vs-bigint-52008685b744
* https://pganalyze.com/blog/5mins-postgres-uuid-vs-serial-primary-keys

Akshat-Jain: Thanks, will check!

extensions/omni_python/docs/intro.md
yrashk: Yeah, it is a minor detail, no perfect answer just yet. I am just urging to think for a minute how others will read this and whether it'll be confusing to them.

Akshat-Jain: In-line logic approach: Seems unnecessarily inconsistent with the remaining endpoints to me.

`employee_to_json()` approach: Would require fetching 0th element and doing `return employees_to_json(list of 1 employee)` == unnecessary back and forth parsing.

That's my opinion on this, though open to any suggestions as always!

PR #332, opened 2023-11-06
Problem: maintaining single .sql migration files

yrashk:
They make it difficult to compose final shippable SQL scripts for every particular version. They also grow out of proportion.

Solution: write them incrementally and compose on psql/test/packaging


yrashk:
For the record, @burntcarrot also suggested this on Discord:

<blockquote>
I was thinking in a similar way, but I had an idea of introducing "block"-based extension creation. Each block will have the procedures related to it with incremental migration. Each extension has a tiny config; I don't have a concrete structure for the config, but it at least will specify what blocks are a part of this extension. And there would be an optional config in the project's root directory, which specifies the available extensions and their path.

A CLI or a script can check these directories (good time to build a omni CLI!), package the migrations to a single SQL file. The artifacts are then either stored in the extensions' directories or it could be stored in a different path.

One advantage of using this approach is that when we develop blocks/extensions which are experimental, we can modify the config to ignore the experimental blocks/extensions when we build for a push on the main branch.
</blockquote>

image

And there was a note that there isn't necessarily a conflict between the ideas (one can grow into another one)

PR #334, opened 2023-11-06
Set Content-Length on client requests regardless of body length

jaggederest:
#279 

Omits the length check for `request->request_body.len` in favor of always setting a `Content-Length` header on requests.

This might more ideally be done with an `else` statement setting it always to zero when `request->request_body.len` is <= zero, but I'm not sure which situations might result in a null or negative content length for the response body, so I merely deleted it for now.

I've found it difficult to grab the content-length header in the tests, would appreciate a pointer, my code might have an error or this may be due too something like the `omni_httpd` response not including it in the headers array as set up in the test, I'm not sure what the expected value of `request.headers` in the `omni_httpd` handler would be if a content-length were set on the request. I think ideally I'd be able to do something like `response.raw_headers` to get access to that but I might be mistaken.


yrashk:
Thank you for your contribution!

> I've found it difficult to grab the content-length header in the tests, would appreciate a pointer

Can you please elaborate on what was the difficulty with it?

jaggederest:
> Thank you for your contribution!
> 
> > I've found it difficult to grab the content-length header in the tests, would appreciate a pointer
> 
> Can you please elaborate on what was the difficulty with it?

Yes, I'm just unfamiliar with the query syntax, I couldn't find any documentation to e.g. specifically select out the `content-length` header from the request-as-returned-in-the-response in the `extensions/omni_httpc/tests/test.yml` tests, to add a check that it is present in the existing tests. So I modified one test (it's failing) as I would expect it to work, and thought we could discuss. Does that clarify?

jaggederest:
> I see. `omni_http` hasn't been documented (and it should have been!) but I wonder if [omni_http.http_header_get](https://github.com/omnigres/omnigres/blob/master/extensions/omni_http/migrate/2_http_headers.sql#L28) will be helpful here?

Yes I think I will need to modify the handler used by `omni_httpd` in the `omni_httpc` test, I will give that a try later today

PR #335, opened 2023-11-06
Problem: omni_containers depends on libcurl & yyjson

### LINKED ISSUE #319: Problem: omni_containers depends on libcurl & yyjson
These dependencies complicate our build unnecessarily, and we miss the benefits of using omni_httpc. Maintaining the C code as it is is becoming burdensome.

Proposed solution: rewrite most of omni_containers in PL/pgSQL

Instead of libcurl & yyjson we can use omni_httpc and native JSON support in Postgres.

Likely, some functionality will remain implemented in C.

In particular, the processing of streams to text has to unpack frames with network byte order headers, for which one would likely need to expose `pg_ntoh32` as a stored procedure (see https://github.com/omnigres/omnigres/blob/master/extensions/omni_containers/omni_containers.c#L409-L428)

However, the rest of the logic can be migrated away.

---

**Bounty size: $250**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

rmodpur:
These dependencies complicate our build unnecessarily, and we miss the benefits of using omni_httpc. Maintaining the C code as it is is becoming burdensome.
    
Solution: port to plpgsql and reuse existing extensions like `omni_httpc` and `omni_web` to reduce dependency.

Closes #319


extensions/omni_httpc/omni_httpc.c
```diff
@@ -317,6 +317,13 @@ static h2o_httpclient_head_cb on_connect(h2o_httpclient_t *client, const char *e
   *_method = req->method;
   *num_headers = req->num_headers;
   *headers = req->headers;
+    /* 
+    required to prevent malformed host header with unix domain sockets
+    for eg. when calling http://[unix:/var/run/docker.sock]/_ping
+    h2o sets the host header to [unix:/var/run/docker.sock] and golang
+    disallows '/' in host header value with 400 Bad request
+  */
+  req->url.authority.len = 0;
```
rmodpur, line 326: This is a hack for now, the eventual purpose is to set host header for unix domain socket connections to empty value. Perhaps `http_execute_with_options` could have a new option for indicating the underlying transport(in this case unix domain socket) to configure `h2o` requests` appropriately

rmodpur:
Some observations:
1. Every once is a while the query takes some time to execute about `4s` but subsequent queries complete in less than `50ms` probably due to `h2o` keeping the conn alive after the initial query
2. Some places have hardcoded values for docker create payload because it depended on env variables, i don't see a way to inspect env variable purely through sql. Either create custom pg settings for such values or inspect it through some c code.

yrashk:
1. That shouldn't happen; if it does (that is, h2o part is responsible, not the docker daemon), we can consider this a buggy behavior and handle accordingly.
2. Two things I can think of: get env variables in C; or rely on GUC (postgres config vars)

rmodpur:
@yrashk I have pushed complete implementation of `docker_container_create`(needs a bit of cleanup but overall structure is there) any thoughts ?
The dependencies will be `omni_web` for url encoding and `omni_httpc` for making requests

yrashk (commented):
Great work! Besides some side issues (like h2o-related stuff), anything holding us from getting tests to run with the new implementations and switching away from C?

rmodpur:
> Great work! Besides some side issues (like h2o-related stuff), anything holding us from getting tests to run with the new implementations and switching away from C?

I'm trying to add `omni_httpc` and `omni_web` to `REQUIRES` part of `omni_containers` in it's `add_postgresql_extension` but cmake config stage is failing with this kind of error
```bash
CMake Error at cmake/PostgreSQLExtension.cmake:146 (add_custom_target):
  add_custom_target cannot create target "omni_http" because another target
  with the same name already exists.  The existing target is a custom target
  created in source directory
  "omnigres/extensions/omni_http".  See documentation for
  policy CMP0002 for more details.
Call Stack (most recent call first):
  extensions/omni_http/CMakeLists.txt:11 (add_postgresql_extension
```

yrashk:
Very interesting!

That line number suggests somewhere there's an extension defined without `SOURCES property:

https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L146

How does your updated CMakeLists.txt look like if you haven't included it in the PR yet? Does this happen with the addition of omni_httpc or omni_web?

rmodpur:
> Very interesting!
> 
> That line number suggests somewhere there's an extension defined without `SOURCES property:
> 
> https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L146
> 
> How does your updated CMakeLists.txt look like if you haven't included it in the PR yet? Does this happen with the addition of omni_httpc or omni_web?

I tried with both individually as well, both failed with similar errors

yrashk:
Can confirm this on master. It's happening because omni_containers is defined before its dependencies.

This is the fault of our current build system:

https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L152-L162

We need to make sure `add_subdirectory` there doesn't get called if it was called already. Perhaps what we can do here is to add our own macro wrapping add_subdirectory that expands to `add_subdirectory` if it hasn't expanded once yet.

yrashk:
Can you try and see if the change in #341 would help?

rmodpur:
> Can you try and see if the change in #341 would help?

Yes, it works now

yrashk:
I've merged #341 into master

yrashk:
I noticed failures with `omni_httpc` tests. I believe #346 will fix this.

extensions/omni_httpc/omni_httpc.c
```diff
@@ -317,6 +317,16 @@ static h2o_httpclient_head_cb on_connect(h2o_httpclient_t *client, const char *e
   *_method = req->method;
   *num_headers = req->num_headers;
   *headers = req->headers;
+  /*
+  required to prevent malformed host header with unix domain sockets
+  for eg. when calling http://[unix:/var/run/docker.sock]/_ping
+  h2o sets the host header to [unix:/var/run/docker.sock] and golang
+  disallows '/' in host header value with 400 Bad request
```
yrashk, line 324: Should we include a reference to a formal (or informal) description of the syntax here or in the docs?

rmodpur: I will add description here, the syntax I believe is h2o specific so i suggest we introduce some sort of abstraction before introducing unix domain sockets to the end user

rmodpur:
macos tests are failing due to hardcoded values in plpgsql instead of using env variables

yrashk:
Great work! Minor nitpick: some of the `Problem` lines in commit messages are formulated through solution. Not a big deal.

yrashk:
I've removed the unused dependencies and fixed a small JSON building issue.

PR #337, opened 2023-11-07
Problem: mkdocs highlighted line range is incorrect

Akshat-Jain:
Solution: Fix the highlighted line range.

Test plan: Verified running mkdocs locally.

Screenshot:
image


yrashk:
Thank you!

PR #338, opened 2023-11-07
Problem: omni_schema may load sources before dependencies are ready

### LINKED ISSUE #327: Problem: omni_schema's order of loading of files in `load_from_fs` is undefined
Proposed solution 1: We can maybe provide an option to configure this. The end goal is to avoid any confusion for the users.

Proposed solution 2: Establish some defined order

Proposed solution 3: combination of the above

yrashk:
Solution: introduce tools/language ordering priority while sorting files by name for consistency

Closes #327

### DIFF
--- a/extensions/omni_schema/migrate/1_languages.sql
+++ b/extensions/omni_schema/migrate/1_languages.sql
@@ -5,7 +5,8 @@ create table languages
     language                 name not null,
     extension                name,
     file_processor           name,
-    file_processor_extension name
+    file_processor_extension name,
+    priority                 integer not null default 0
 );
 
 insert
--- a/extensions/omni_schema/migrate/2_aux_tools.sql
+++ b/extensions/omni_schema/migrate/2_aux_tools.sql
@@ -4,13 +4,14 @@ create table auxiliary_tools
     filename_stem       text,
     filename_extension  text,
     processor           name not null,
-    processor_extension name
+    processor_extension name,
+    priority            integer not null default 0
 );
 
 insert
 into
-    auxiliary_tools (filename_stem, filename_extension, processor, processor_extension)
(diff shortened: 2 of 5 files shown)


extensions/omni_schema/migrate/1_languages.sql
```diff
@@ -5,7 +5,8 @@ create table languages
     language                 name not null,
     extension                name,
     file_processor           name,
-    file_processor_extension name
+    file_processor_extension name,
+    priority                 integer not null default 0
```
Akshat-Jain, line 9: I like the implementation idea of this table!

Also, yay, I know one more table now! :D

extensions/omni_schema/migrate/2_aux_tools.sql
```diff
@@ -4,13 +4,14 @@ create table auxiliary_tools
...
+    priority            integer not null default 0
 );
 
 insert
 into
-    auxiliary_tools (filename_stem, filename_extension, processor, processor_extension)
+    auxiliary_tools (filename_stem, filename_extension, processor, processor_extension, priority)
 values
-    ('requirements', 'txt', 'omni_python.install_requirements', 'omni_python');
+    ('requirements', 'txt', 'omni_python.install_requirements', 'omni_python', 1);
```
Akshat-Jain, line 15: Question: Do we need order when we load multiple Python files that have some dependency between them? If yes, this approach doesn't handle that (unless I'm missing something!).
Appreciate your thoughts!

yrashk: This approach doesn't handle this case at all. I just wanted to establish some predictable and sane order.

The dependency mechanism would likely involve custom importers and is more on omni_python side than omni_schema:

https://docs.python.org/3/library/importlib.html

Akshat-Jain (approved):
Looks good!

PR #342, opened 2023-11-09
Problem: Python code importing local modules

yrashk:
omni_schema + omni_python do not collaborate to provide local imports. One has to resort to providing path-based dependencies through requirements.txt.

Solution: implement a custom importer

The importer requires a (currently janky [1]) integration with omni_vfs.

[1] We can't pass `anyelement` in plpython so we're resorting to pass the textual representation of the fs and its type and then sticking that into a new query. Would having our own fork of plpython resolve this?

--- 

Action plan:

- [x] The importer would only consider files in the same directory, no submodules or directories. This is just to get going. This has to be fixed.

### DIFF
--- a/extensions/omni_python/migrate/2_functions.sql
+++ b/extensions/omni_python/migrate/2_functions.sql
@@ -42,16 +42,20 @@ $$
         if have_omni_vfs:
             import importlib.abc, importlib.machinery
             class CustomImporter(importlib.abc.MetaPathFinder, importlib.abc.Loader):
+                query = f"""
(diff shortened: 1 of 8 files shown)


yrashk:
Currently, CI builds fail with this error:

```
ModuleNotFoundError: No module named 'module1.submod'
```

I had this error up until I used `is_package=True`. CI runs Python 3.10.12. I typically run dev env on 3.11. However, I retried on python3.10.12 and can't reproduce the failure.

PR #345, opened 2023-11-09
Problem: fast is never fast enough

yrashk:
We all want omni_httpd to be as fast as possible

Solution: mark commonly used functions immutable as an optimization

Measured a ~3% improvement on a simple static response test.

### DIFF
--- a/extensions/omni_httpd/migrate/1_basic_types.sql
+++ b/extensions/omni_httpd/migrate/1_basic_types.sql
@@ -14,6 +14,7 @@ create function http_request(path text, method omni_http.http_method default 'GE
                              headers http_headers default array []::http_headers)
     returns http_request
     language sql
+    immutable
 as
 $$
 select row (method, path, query_string, body, headers)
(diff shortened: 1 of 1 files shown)


JoshInnis:
Marking them as STABLE might also yield the same performance gain.  IMMUTABLE states that the function will always return the same result every time the function gets run, if the input doesn't change. VOLATILE, which is the default says it can change on a per call basis. STABLE says the result won't change in the given transaction you are in but may change on a query-by-query basis. 

If you make them STABLE, you can't include function in an evaluated expression, however, if the arguments are constant in a query, Postgres will call the function once per query and cache the results, and you should see similar performance benefits , but you avoid people being able to use the functions in Indices or other places where they will not work correctly.

yrashk:
@JoshInnis thank you!

Yes, I am aware of the distinction, which is quite critical. The reason I marked these functions immutable is because I believe that they do, in fact, adhere to the immutable properties. These functions assemble composite (and sum) types from provided values. I can, however, be wrong. A particularly important thing to be aware of here is that when we change (upgrade) any of the types used by the function, we need to reload function definitions.

Outside of that, the actual endpoint functions are not marked stable or volatile, of course.

## Performance

I did a few non-scientific tests to compare immutable and stable.

### Immutable

```sql
select omni_httpd.http_response('hello') from request
```

I tested with:
```
oha -z 10s -c 200 http://127.0.0.1:8080
```

and got **151207.4316** requests/sec.

I also changed the implementation to

```sql
select omni_httpd.http_response('hello' || random()) from request
```

and got **140575.984** requests/sec. I verified that it returns different values, as expected:

```
~ $ curl http://127.0.0.1:8080/
hello0.5403533062660588⏎
~ $ curl http://127.0.0.1:8080/
hello0.361827162732445⏎
~ $ curl http://127.0.0.1:8080/
hello0.434149075877055⏎
~ $ curl http://127.0.0.1:8080/
hello0.04605758154375916⏎
~ $ curl http://127.0.0.1:8080/
hello0.4937873152866734⏎
```

### Stable

On the static endpoint, I got **142711.1759** requests/sec and on the`random()` one I got **141607.7264** requests/sec

PR #360, opened 2023-11-13
Problem: Duplicate migration versions are not restricted.

### LINKED ISSUE #348: Problem: Duplicate migration versioning is not restricted
Right now, we do not restrict having multiple migrations on the same version. Ideally, this should be restricted.

Akshat-Jain:
Solution: Detect duplicate migration versions and throw error accordingly.

Closes #348

### DIFF
--- a/cmake/PostgreSQLExtension.cmake
+++ b/cmake/PostgreSQLExtension.cmake
@@ -318,8 +318,22 @@ if [ -d \"$_dir/${NAME}\" ]; then
   _dir=\"$_dir/${NAME}\"
 fi
 # Create file (using $$ for pid to avoid race conditions)
+
 for f in $(ls \"$_dir/\"*.sql | sort -V)
   do
+  filename=$(basename \"$f\")
+  if [[ $filename =~ ([0-9]+)_ ]]; then
+    migration_number=\${BASH_REMATCH[1]}
+    if grep -q \" $migration_number \" <<< \"$unique_versions\"; then
+      echo \"Error: Duplicate migration number found: $migration_number in filename: $filename\" >&2
+      exit 1
+    else
+      unique_versions+=\" $migration_number \"
+    fi
+  else
+    echo \"Error: Unable to extract migration version from filename: $filename\" >&2
+    exit 1
+  fi
   $<TARGET_FILE:inja> \"$f\" >> \"$1/_$$_${NAME}--${_ext_VERSION}.sql\"
   echo >> \"$1/_$$_${NAME}--${_ext_VERSION}.sql\"
 done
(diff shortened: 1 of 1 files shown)


cmake/PostgreSQLExtension.cmake
```diff
@@ -318,16 +318,30 @@ if [ -d \"$_dir/${NAME}\" ]; then
...
 # Create file (using $$ for pid to avoid race conditions)
+
 for f in $(ls \"$_dir/\"*.sql | sort -V)
   do
+  filename=$(basename \"$f\")
+  if [[ $filename =~ ([0-9]+)_ ]]; then
+    migration_number=\${BASH_REMATCH[1]}
+    if grep -q \" $migration_number \" <<< \"$unique_versions\"; then
+      echo \"Error: Duplicate migration number found: $migration_number in filename: $filename\" >&2
+      exit 1
```
Akshat-Jain, line 329: @yrashk While the logic works and it throws error code 1 from the generated scripts (`script_omni_python` for example), this is not error-ing out the `make` command.

Can you please guide me on how to achieve that? Thanks!

cmake/PostgreSQLExtension.cmake
```diff
@@ -318,16 +318,30 @@ if [ -d \"$_dir/${NAME}\" ]; then
...
+    else
+      unique_versions+=\" $migration_number \"
+    fi
+  else
+    echo \"Error: Unable to extract migration version from filename: $filename\" >&2
```
Akshat-Jain, line 334: I just realised that we have some migration files that don't follow this naming scheme, like `libpgaug_test.sql`.

Is that intentioned? Should I go ahead and rename those?

yrashk: Unintentional, renaming is good!

cmake/PostgreSQLExtension.cmake
yrashk: I apologize for forgetting about this, slipped through the cracks.

Which `make` target does it not fail on? `make psql_`, or tests, I presume?

My intuition here suggests that it has to do with how other scripts handle failure of `script_EXT`.

What I can note here:

* `script_` itself simply skips dependant `script_` invocations when they fail: https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L325
* pg_regress test harness doesn't check for the errors: https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L419
* `psql_` doesn't either: https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L484
* pg_yregress harness _should_ work but worth investigating: https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L122

Akshat-Jain:
I'm not actively working on this anymore, so I'll close off this PR.

PR #363, opened 2023-11-13
Problem: docker-related tests failing on macOS in GitHub Actions

yrashk:
Error messages suggest that the IP address set up in DOCKER_HOST_IP is not correct.

Solution: replace hard-coded IP address with a script


yrashk:
Not really fully fixed yet, but another issue (first byte read timeout) will be addressed in a different PR.

Issue #315, opened 2023-10-30
Problem: intermittent CI problem with http2_reproxy


yrashk:
This is the test: https://github.com/omnigres/omnigres/blob/master/extensions/omni_httpd/tests/http2_reproxy.yml

It often fails due to a timeout in CI, and we have to re-run it.

The behaviour it is testing is as follows:

Considering that we have only two workers (https://github.com/omnigres/omnigres/blob/master/extensions/omni_httpd/tests/http2_reproxy.yml#L6)

If we send two HTTP/2 requests over the same connection, instead of blocking on the second request (since the Postgres worker is busy), omni_httpd will reproxy the request to the other available worker. 

It is tested by making the backend sleep for 2  seconds. The test ensures that 
1) it takes no longer than 2.5 seconds total 
2) responses return different pids for their backends.

I wonder if this can be tested better (preferably without timeouts involved) or if the implementation is deficient. This feature has undergone manual testing before using a command line HTTP client to observe the behaviour.

---

**Bounty size: $100**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

rmodpur:
From the [postgres documentation](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-DELAY)
>The sleep delay will be at least as long as specified. It might be longer depending on factors such as server load.

It can simply be the case that `pg_sleep(2)` is taking longer time to finish occasionally leading to timeout.

yrashk:
Indeed, it is a likely cause. Thinking if we can find a better test for this behavior.

gouravmpk:
Assign

yrashk:
#414 seemingly solves this problem. We can reopen if it comes back.

rmodpur:
It is not exactly reproducible but sometimes it does fail with `error: first byte timeout`

PR #375, opened 2023-11-20
Update http2_reproxy.yml

gouravmpk:
intermittent CI problem with http2_reproxy #315


yrashk:
The problem has not been observed in a while.

PR #382, opened 2023-11-26
Use extension name for tests

### LINKED ISSUE #381: Problem: Impossible to keep tests for different extensions in the same directory
At the moment, pg_yregress expects to find tests in the `tests` folder. We also scan the test folder only once when looking for tests. This makes it impossible to have tests for different extensions in the same tests folder since we end up using the same setup script for both extensions.

## Proposed Solution
During scanning, we set a "cache"/global property to mark the directory as "scanned" https://github.com/omnigres/omnigres/blob/761446d11ccd775f32ab7026c4ac4b9041b60ecb/cmake/PostgreSQLExtension.cmake#L128

We can extend this cache key to include the extension name:
`set_property(GLOBAL PROPERTY "find_pg_yregress_tests:${dir}:${NAME}" "CALLED")`

idoqo:
At the moment, pg_yregress expects to find tests in the tests folder. We also scan the test folder only once when looking for tests. This makes it impossible to have tests for different extensions in the same tests folder since we end up using the same setup script for both extensions.

This adds support for multi-extension directories, and we can now have `tests/extension-name/tests.yml`. Since most of the current tests are for *only one* extension, I figured it makes sense to keep the current behavior of scanning just the `tests` directory. 

Closes #381


yrashk:
Great contribution, thank you! In the future, please consider using Problem/Solution statements in your commit messages for clarity.

PR #384, opened 2023-11-27
Problem: passing fds to http_worker can be fragile

lkshminarayanan:
The transfer of fds is limited to SCM_MAX_FDs and might be fragile.

Solution: Add support for passing more than SCM_MAX_FDs and also use correct buffer size for the FD_BUFFER macro.

Summary of changes and other analysis w.r.to the quirks mentioned in the [gist](https://gist.github.com/kentonv/bc7592af98c68ba2738f4436920868dc) :

1. Updated [FD_BUFFER](https://github.com/omnigres/omnigres/blob/761446d11ccd775f32ab7026c4ac4b9041b60ecb/extensions/omni_httpd/fd.c#L30) definition :
  The current buffer size is unnecessarily large : `n * (sizeof(int) + sizeof (struct cmsghdr))`. 
  Changed it to `(n * sizeof(int)) + sizeof (struct cmsghdr)`.

2. Added support for passing more than MAX_N_FDS. The fds will be split up and sent in multiple messages if the count is more than MAX_N_FDS.

3. [recv_fds()](https://github.com/omnigres/omnigres/blob/761446d11ccd775f32ab7026c4ac4b9041b60ecb/extensions/omni_httpd/fd.c#L101) always provide adequate buffer size, so the ancillary message will never be truncated and there is no risk of open fds leaking without being closed.

4. `cmsghdr.cmsg_len` is always less that `msghdr.msg_controllen` in the current implementation - so there is no risk of former overrunning the later triggering a segfault.

### DIFF
--- a/extensions/omni_httpd/fd.c
+++ b/extensions/omni_httpd/fd.c
@@ -9,11 +9,11 @@
 #define _XOPEN_SOURCE_EXTENDED 1
 #endif
 
-#include <assert.h>
 #include <stdlib.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/uio.h>
+#include <unistd.h>
 #if defined(__FreeBSD__)
 #include <sys/param.h>
 #endif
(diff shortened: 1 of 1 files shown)


lkshminarayanan:
I have manually tested and verified that the code works but there is no test case yet - I'm not sure how/where to write one.

yrashk:
@lkshminarayanan, thank you for sending this PR! Looks like the CI results in some segfaults across Linux boxes. Do you have any thoughts on this?

lkshminarayanan:
@yrashk My usage of  `CMSG_FIRSTHDR` before initialising `msghdr.msg_controllen` was the problem. I have now updated the patch fix the issue.

yrashk:
Apologies for the delay in reviewing this.

With regards to the "quirks gist", are there any quirks there in the gist or in the comments that may be applicable to our case that were not addressed in this PR?

Ideally, I am trying to get to a summary table of "Issue raised in the gist" and "Action taken" (none [and why], some [details]) to make sure we have captured the knowledge here.

lkshminarayanan:
@yrashk Sure! Here is a list of items and action taken :

> 1. recvmsg() calls on the receiving end don't necessarily map 1:1 to sendmsg() calls. Messages can be coalesced or split.

- No action required. 
- Messages are split only when the receiver fails to provide a sufficient buffer, as demonstrated in this  [stack exchange question](https://unix.stackexchange.com/q/185011). The function `read_fds_with_buffer` provides an ample buffer avoiding any splits.
- Two messages are coalesced only if one of them has an an ancillary message or neither has any ancillary message. In the case of `send_fds_with_buffer`,  all messages have an ancillary messages attached, so they do not get coalesced.

> 2. The recvmsg() call that receives the first byte of the ancillary message's byte range also receives the ancillary message itself.

- No action required
- `read_fds_with_buffer` always provides an adequate buffer ensuring that the entire message is received along with the ancillary message which is attached to the first byte of the real message. 

> 3.
> - To prevent multiple ancillary messages being delivered at once, the recvmsg() call that receives the ancillary data will be artifically limited to read no further than the last byte in the range, even if more data is available in the buffer after that byte, and even if that later data is not actually associated with any ancillary message.
> - However, if the recvmsg() that received the first byte does not provide enough buffer space to read the whole message, the next recvmsg() will be allowed to read past the end of the mesage range and even into a new ancillary message's range, returning the ancillary data for the later message.

- Doesn't apply here as the messages are always read 1:1.

> 4. If the recvmsg() does not provide enough ancillary buffer space to fit the whole descriptor array, it will be truncated to fit, with the remaining descriptors being discarded and closed.

- Doesn't apply as an adequate buffer is always provided.

> 5. The byte range that the ancillary message is attached to cannot have zero size.
- `send_fds_with_buffer`  payload is always greater than zero.

> 6. If you call recvmsg() and provide space to receive an ancillary message at all -- even if you DIDN'T expect SCM_RIGHTS -- you MUST check if you received an SCM_RIGHTS message and, if so, close the file descriptors. Otherwise, an attacker can fill up your file descriptor table with garbage, probably DoSing you.
-  `recv_fds_with_buffer` always expects SCM_RIGHTS and the received fds are always handled appropriately. 

> 7. A single recvmsg() can in fact receive multiple messages, any of which could be SCM_RIGHTS. Don't forget to check for this.
- Since all messages have ancillary messages attached and since `recv_fds_with_buffer` always provides an adequate buffer, `recvmsg()` can never receive multiple messages in this current implementation.

> 8. Some operating systems are buggy in the case that the ancillary buffer space is too small to fit all received descriptors. Notably, on MacOS, the excess descriptors that were never delivered to the process via the ancillary message will nevertheless have been added to the process's file descriptor table, and so will not be closed. Until very recently, [FreeBSD had this bug as well](https://bugs.freebsd.org/131876). In these cases, in order to avoid DoS attacks, it is necessary to provide a buffer that is larger than the maximum number of FDs that can be transferred at once. Unfortunately, this means you must temporarily accept all these FDs into your FD table, and then go close them. Also, the maximum number of FDs is not documented.
- Not an issue as, again, adequate buffer size is used by `recv_fds_with_buffer`.

> 9. Additionally, on MacOS, cmsghdr.cmsg_len is allowed to overrun the underlying buffer space indicated by msghdr.msg_controllen. You must check for this and clamp otherwise you are likely to overrun the buffer. In the best case, you segfault. In the more-likely case, you end up calling close(0), which leads to madness. On Linux, the kernel automatically clamps cmsg_len so that it does not overrun.
- `cmsghdr.cmsg_len` is always less than `msghdr.msg_controllen` in the current implementation - so there is no risk of former overrunning the later triggering a segfault.

Let me know if you need more details.

PR #386, opened 2023-11-27
Problem: sporadic failures in macOS CI

yrashk:
The error looks like this:

```
dyld[62749]: Library not loaded: /Users/ci/actions-runner/_work/omnigres/omnigres/.pg/Darwin-Release/16.1/build/lib/libpq.5.dylib
  Referenced from: <9334436C-BA46-32DC-873D-CAAE0C09F2A1>
  /Users/ci/actions-runner1/_work/omnigres/omnigres/build/pg_yregress/pg_yregress
```

Solution: ensure we use correct install names of libraries

Bascially, we have two action runners, in $HOME/actions_runner and $HOME/actions_runner1 and when they get .pg restored from the action cache, the libraries that get built in one (like libpq) have install_name that contains a full path to where it was built. Thus. When we have `actions_runner/_work/omnigres/omnigres/.pg/.../libpq.dylib` and then restore it as
`actions_runner1/_work/omnigres/omnigres/.pg/.../libpq.dylib` and link against it, at the start time it'll try to load install_name (the original path) and will fail.

So, for now, we'll just make sure that all binaries and libraries are linked against correct libpq.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,13 @@ jobs:
         path: ${{github.workspace}}/build/_deps
         key: ${{ github.workflow }}-cpm-modules-${{ hashFiles('extensions/**/CMakeLists.txt', '*/CMakeLists.txt', 'cmake/*.cmake') }}
 
+    # This is to address the fact that we build .pg in different directories on our current macOS runners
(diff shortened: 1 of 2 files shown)


yrashk:
Resolved differently in #389

Issue #387, opened 2023-11-29
Problem: Brew updated OpenSSL formulae. Breaks our build on MacOS


viggy28:
OpenSSL version to search for on MacOS is hardcoded to 3.1. Looks like OpenSSL released a new version 3.2 (https://github.com/openssl/openssl/releases) and as part of that the formulae have been updated. Hence, `COMMAND brew --prefix openssl@3.1` no longer resolves properly.

When trying to build, it fails

```
Error: No available formula with the name "openssl@3.1". Did you mean openssl@3.0, openssl@3 or openssl@1.1?
CMake Error at cmake/OpenSSL.cmake:8 (message):
  No OpenSSL found, use homebrew to install one
Call Stack (most recent call first):
  cmake/FindPostgreSQL.cmake:54 (include)
  pg_yregress/CMakeLists.txt:11 (find_package)
```

Solution: Update the formulae version to search as 3.

PR #388, opened 2023-11-29
FIX: update openssl version to latest major release which is 3

viggy28:
Fixes https://github.com/omnigres/omnigres/issues/387

### DIFF
--- a/cmake/OpenSSL.cmake
+++ b/cmake/OpenSSL.cmake
@@ -1,6 +1,6 @@
 if(NOT DEFINED OPENSSL_CONFIGURED)
     if(APPLE)
-        execute_process(COMMAND brew --prefix openssl@3.1
+        execute_process(COMMAND pkg-config --variable=prefix openssl
                 OUTPUT_VARIABLE OPENSSL_PREIX RESULT_VARIABLE OPENSSL_RC
                 OUTPUT_STRIP_TRAILING_WHITESPACE)
 


cmake/OpenSSL.cmake
```diff
@@ -1,15 +1,34 @@
...
-        set(OPENSSL_ROOT_DIR ${OPENSSL_PREIX} CACHE INTERNAL "OpenSSL")
+        is_symlink(${OPENSSL_PREFIX} is_symlink_result)
+        message(STATUS "Found OpenSSL at ${OPENSSL_PREFIX}")
+        if(is_symlink_result)
+            file(REAL_PATH ${OPENSSL_PREFIX} absolute_path)
+            set(OPENSSL_ROOT_DIR ${absolute_path} CACHE INTERNAL "OpenSSL")
+        else()
+            set(OPENSSL_ROOT_DIR ${OPENSSL_PREFIX} CACHE INTERNAL "OpenSSL")
+        endif()
+        message(STATUS " OpenSSL root dir at ${OPENSSL_ROOT_DIR}")
```
yrashk, line 31: Any particular reason to save the absolute path? Is it so that if it changes over time (homebrew updates openssl@3 to point to 3.3 or whatever), we still use the version we picked previously?

Would be prudent to document the reasoning here.

cmake/OpenSSL.cmake
```diff
@@ -1,15 +1,34 @@
...
+        set(${result} FALSE PARENT_SCOPE)
+    endif()
+endfunction()
+
 if(NOT DEFINED OPENSSL_CONFIGURED)
     if(APPLE)
-        execute_process(COMMAND brew --prefix openssl@3.1
-                OUTPUT_VARIABLE OPENSSL_PREIX RESULT_VARIABLE OPENSSL_RC
+        execute_process(COMMAND brew --prefix openssl@3
+                OUTPUT_VARIABLE OPENSSL_PREFIX RESULT_VARIABLE OPENSSL_RC
```
yrashk, line 17: This has a potential pitfall of being too low (3.0); figuring this out may involve calling and parsing the output of `brew info --json openssl@3` to see what actual version did it resolve to.

cmake/OpenSSL.cmake
```diff
@@ -1,15 +1,34 @@
...
 
         if(NOT OPENSSL_RC EQUAL 0)
             message(FATAL_ERROR "No OpenSSL found, use homebrew to install one")
         endif()
-
-        message(STATUS "Found OpenSSL at ${OPENSSL_PREIX}")
-        set(OPENSSL_ROOT_DIR ${OPENSSL_PREIX} CACHE INTERNAL "OpenSSL")
+        is_symlink(${OPENSSL_PREFIX} is_symlink_result)
+        message(STATUS "Found OpenSSL at ${OPENSSL_PREFIX}")
+        if(is_symlink_result)
```
yrashk, line 25: cmake has native "is_symlink" functionality: https://cmake.org/cmake/help/latest/command/if.html#is-symlink

yrashk (changes requested):
Overall reasonable, but a few inline comments / requests for improvement.

Also to note:

1. The commit message is not formulated through Problem/Solution, which is the convention. Issue #387 is doing a good job of this, though.
2. The commit message may be somewhat misleading as it does not indicate that it it to use the latest minor version of the major release 3.

cmake/OpenSSL.cmake
yrashk: Also, thinking of this a bit deeper, it might just make sense always to get the absolute path, whether it is a symlink or not, making the code a lot simpler.

cmake/OpenSSL.cmake
viggy28: Fixed through the latest commit.

cmake/OpenSSL.cmake
viggy28: I overlooked it. No need to cache that value. Removed in the latest commit.

yrashk:
Superseded by #390

PR #390, opened 2023-11-29
Problem: finding the correct version of OpenSSL

yrashk:
This is often a problem as target systems may have a wrong version, and discovery methods vary between systems and installation methods.

Solution: bring OpenSSL build to the configure stage if necessary

If a correct version of OpenSSL is not found, let's build it.

This also builds Perl to avoid depending on system Perl and Text::Template. On its own this seems a bit ridiculous, but considering we will be building PL/Perl, this would make more sense in the future.

This increases the configure time but (hopefully) makes builds more reliable.


rmodpur:
This could help https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

yrashk:
> https://cmake.org/cmake/help/latest/module/FindOpenSSL.html

Finding the correct version only works if the target system has the correct version :) There are also systems where those installations don't have static versions of the library (which we use)

rmodpur:
Yeah it won't install it for the user but will give an useful error and point out what kind of packages are missing from their system

yrashk:
I've updated this PR to only build from scratch if a correct version can't be found.

Issue #392, opened 2023-11-30
Problem: Build extensions separately is currently failing


viggy28:
We have a feature to build extensions separately. However, it's failing since individual extensions don't include some of the dependencies. For. eg `omni_http` requires `inja` and `omni_ext`. When we are building from the root directory, CMakeList.txt includes all of the necessary dependency targets. Making every file (`FindPostgreSQL`, `PostgreSQLExtension` etc) support building from both the root directory and individual directory makes the logic more complex.

Solution: Add the necessary targets as `DEPENDS_ON`. It's a bit of redundancy but it at least avoids the complexity.

NOTE: Regarding why the CI is not currently failing on this, `build-extensions-separately` script doesn't check for errors on any of the `cmake` commands. One solution is to add `set -e`, so a failure of any of the commands on the script will fail the job.

PR #393, opened 2023-11-30
Problem: Build extensions separately is currently failing

viggy28:
Fixes https://github.com/omnigres/omnigres/issues/392

Because individual extensions don't include some of the dependencies. For. eg `omni_http` requires `inja` and `omni_ext`.

Solution: add the necessary dependencies through `DEPENDS_ON` argument.


yrashk:
Perhaps we can make inja a requirement added throughout every extension in Postgre without spelling it out in every extension? This way it'll be less error-prone.

cmake/FindPostgreSQL.cmake
```diff
@@ -51,7 +51,11 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-include(${CMAKE_SOURCE_DIR}/cmake/OpenSSL.cmake)
+include(${CMAKE_SOURCE_DIR}/cmake/OpenSSL.cmake OPTIONAL RESULT_VARIABLE file_found)
+
+if(file_found==NOTFOUND) 
+include(${CMAKE_SOURCE_DIR}/../../cmake/OpenSSL.cmake RESULT_VARIABLE file_found)
+endif()
```
yrashk, line 58: Not supper happy about this, need to find a better way. This was previously handled by having something like this:

```CMake
list(PREPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/../../cmake)
```

in the extensions, as it's basically pointing to the right directory. However, ${CMAKE_SOURCE_DIR} in FindPostgreSQL.cmake is bypassing this. I think we need to figure out how we can make it `include(OpenSSL)` again.

viggy28: Updated using `CMAKE_CURRENT_LIST_FILE` - which provides the current CMAKE file's location. Irrespective of from where `FindPostgreSQL.cmake` is called, it includes the `OpenSSL.cmake` file

viggy28:
You are right. I thought it would make the logic more complex, but it isn't. Added `inja` and `omni_ext` as a global dependency.

cmake/FindPostgreSQL.cmake
```diff
@@ -51,7 +51,9 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-include(${CMAKE_SOURCE_DIR}/cmake/OpenSSL.cmake)
+get_filename_component(PRESENT_DIR "${CMAKE_CURRENT_LIST_FILE}" DIRECTORY)
```
yrashk, line 54: Can this be achieved with [CMAKE_CURRENT_LIST_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_DIR.html)

viggy28: Since `FindPostgreSQL.cmake` is called by `CMakeLists.txt` files located in different directories (root, root/extensions/*), `CMAKE_CURRENT_LIST_DIR` will provide the corresponding directory location.  

No matter which CMakeLists.txt file calls `FindPostgreSQL.cmake`, we need to include `OpenSSL.cmake` and we know where it is located. So, I am not sure why we would use `CMAKE_CURRENT_LIST_DIR` for this use case.

cmake/PostgreSQLExtension.cmake
```diff
@@ -171,6 +183,8 @@ function(add_postgresql_extension NAME)
                 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../${dependency}" "${CMAKE_CURRENT_BINARY_DIR}/${dependency}")
             elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../../${dependency}")
                 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../../${dependency}" "${CMAKE_CURRENT_BINARY_DIR}/${dependency}")
+            elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../../misc/${dependency}")
```
yrashk, line 186: Do we still need this with `ninja` not being an explicit dependency anymore?

cmake/PostgreSQLExtension.cmake
```diff
@@ -151,6 +151,18 @@ function(add_postgresql_extension NAME)
...
     add_dependencies(${NAME} check_git_commit_hash)
 
+    # inja ia a default target dependency for all extensions    
+    if(NOT TARGET inja)
+        add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/../../misc/inja" "${CMAKE_CURRENT_BINARY_DIR}/inja")
+    endif()
+
+    # omni_ext is required by all other extensions except omni_ext itself
+    if(NOT ${NAME} STREQUAL "omni_ext")
+        if(NOT TARGET omni_ext)
```
yrashk, line 161: Nitpick: can use `AND` in the `if` to simplify

cmake/FindPostgreSQL.cmake
yrashk: So I applied this change to the current PR:

```diff
diff --git a/cmake/FindPostgreSQL.cmake b/cmake/FindPostgreSQL.cmake
index 6b37a1a1f..feb9cc339 100644
--- a/cmake/FindPostgreSQL.cmake
+++ b/cmake/FindPostgreSQL.cmake
@@ -54,6 +54,9 @@
 get_filename_component(PRESENT_DIR "${CMAKE_CURRENT_LIST_FILE}" DIRECTORY)
 
 include(${PRESENT_DIR}/cmake/OpenSSL.cmake OPTIONAL RESULT_VARIABLE file_found)
+if(NOT ${PRESENT_DIR} STREQUAL ${CMAKE_CURRENT_LIST_DIR})
+    message(FATAL_ERROR "${PRESENT_DIR} but ${CMAKE_CURRENT_LIST_DIR}")
+endif()
 
 if(NOT DEFINED PG_CONFIG)
 
```

Then, I ran `make` for a global build as well as `ci/build-extensions-separately`, and none of them failed with the above error. 

To me, this suggests that `CMAKE_CURRENT_LIST_DIR` is good here. Am I missing anything you might have found?

cmake/PostgreSQLExtension.cmake
yrashk: Also, technically, it's not required by all extensions. I think there was logic around that:

https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L123
https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L395-L397
https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L490

I *suspect* that the problem you may have run into is around this:

https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L509

Where we add a reference to a target whether it exists or doesn't. And we do this for `psql_` commands. I think the main intention behind that was so that we can load any extension from any `psql_` command, including those that require `omni_ext`.

cmake/FindPostgreSQL.cmake
viggy28: You are right. I think I got confused with the definition of `CMAKE_CURRENT_LIST_DIR`. Updated using it.

cmake/PostgreSQLExtension.cmake
viggy28: No, we don't need. Dropped that `elseif`.

cmake/PostgreSQLExtension.cmake
viggy28: Updated.

> I suspect that the problem you may have run into is around this:
That's exactly where I ran into the problem.

cmake/FindPostgreSQL.cmake
```diff
@@ -51,7 +51,11 @@
 # License for the specific language governing permissions and limitations under
 # the License.
 
-include(${CMAKE_SOURCE_DIR}/cmake/OpenSSL.cmake)
+include(${CMAKE_CURRENT_LIST_DIR}/OpenSSL.cmake RESULT_VARIABLE file_found)
+
+if(file_found STREQUAL "NOTFOUND")
+    message(FATAL_ERROR "cant find openssl.cmake file. It is expected to be under cmake directory")
```
yrashk, line 57: Do we need this if `include` itself can fail?

yrashk:
#413 might be tangentially relevant here as it [fixes dynpgext test dependencies](https://github.com/omnigres/omnigres/pull/413/files#diff-76c7f584c23868a916cbfa5fc3dda1fc4a7d5a25a90b8e0b519081c8342b0736)

viggy28:
#413 was relevant. I didn't include the automatic inclusion of new extensions in this PR. 

This PR is ready for review. As it is, I am able to build extensions independently and all together.

cmake/PostgreSQLExtension.cmake
```diff
@@ -394,6 +405,7 @@ $command $@
 
                 if(req STREQUAL "omni_ext")
                     set(_extra_config "shared_preload_libraries=\\\'$<$<TARGET_EXISTS:omni_ext>:$<TARGET_FILE:omni_ext>>\\\'")
+                    message(STATUS "extra_config ${_extra_config}")
```
yrashk, line 408: Debugging message?

cmake/FindPostgreSQL.cmake
yrashk: Does this particular change contribute to fixing the issue? If not, I suggest leaving it out and handling failed includes (any) more generically.

viggy28: Yes, this change does fix the current issue.

yrashk: I didn't mean the `include` itself but the error reporting. Given that we do so many imports everywhere, handling it just here this way doesn't seem to be right.

yrashk: I assume `include` does "send_error" which doesn't make cmake fail right away, but prints the error and exits with an error.

You're just handling the case by converting that "send_error" to "fatal_error"

I suggest we extract the discussion whether file inclusion should be a fatal error or not into a separate issue and come up with a universal solution instead of an ad-hoc handler.

dynpgext/test/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ add_postgresql_extension(
         VERSION 0.1
         SCRIPTS dynpgext_test--0.1.sql
         PRIVATE true
+        REQUIRES omni_ext
```
yrashk, line 18: Actually, I don't think it does. The [whole test](https://github.com/omnigres/omnigres/blob/master/dynpgext/test/tests/loader_present.yml) is just testing the lack of presence of the loader

viggy28: Updated.

cmake/FindPostgreSQL.cmake
viggy28: Alright. Updated to remove the error handling.

Issue #391, opened 2023-11-30
Problem: omni_aws DigitalOcean S3 support


yrashk:
Seemingly everything (with the exception signed) URLs doesn't work with the `digitalocean_s3_endpoint()` backend, returning a signature mismatch error.

---

**Bounty size: $100**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

rmodpur:
From https://docs.digitalocean.com/reference/api/spaces-api/#list-a-buckets-contents

> The version 2 list type is not currently supported.

rmodpur:
Some observations:
1. create bucket api => able to create buckets through extension if payload is empty, digitalocean is rejecting bucket create requests with payload (even with official aws-sdk )
2. put object api => able to put objects through official aws-sdk but is failing with `400Bad Request` for extension

On debug logging official aws sdk i found digitalocean spaces api is responding with HTTP/2.0 response for HTTP/1.1 requests, so i tried switching to http2 for extension but it is failing with generic 'I/O error'

yrashk:
Addressed in #397

PR #397, opened 2023-12-02
Problem: omni_aws DigitalOcean S3 support

rmodpur:
Seemingly everything (with the exception signed) URLs doesn't work with the digitalocean_s3_endpoint() backend, returning a signature mismatch error

Solution: set null path to '/' for methods with optional paths and send empty payload for create bucket because DigitalOcean rejects it

#391

### DIFF
--- a/extensions/omni_aws/migrate/4_s3_list_objects_v2.sql
+++ b/extensions/omni_aws/migrate/4_s3_list_objects_v2.sql
@@ -107,7 +107,8 @@ begin
             omni_aws.endpoint_url(endpoint, bucket => request.bucket, region => region, path => request.path);
     endpoint_uri := omni_web.text_to_uri(endpoint_url);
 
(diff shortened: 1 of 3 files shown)


rmodpur:
`s3_create_bucket`(with no payload), `s3_put_object` and `s3_list_objects_v2` is working with digitalocean spaces with the changes

yrashk:
Do you think may require re-testing against AWS S3?

rmodpur:
> Do you think may require re-testing against AWS S3?

Yeah better to test it before merging, also would be good to have some sort of automated periodic tests(eg. every 2 weeks) against s3 and spaces

rmodpur:
I have tested all the methods with both s3 and spaces, all are working as expected

yrashk:
Excellent, thank you!

Issue #351, opened 2023-11-10
Problem: impossible to run tests in a different database


yrashk:
In some obscure cases (like omni_ext), we want to be able to test against multiple databases in the same instance. For example, some extensions may have database-specific results.

Currently, we can't do `create database` because we're wrapping everything into a transaction (`CREATE DATABASE cannot run inside a transaction block` error), and we can't switch databases on the fly (pg_yregress doesn't have support for it.

To address this, we can perhaps add the following functionality:

## Non-transactional tests in pg_yregress

Something like this?

```yaml
- query: create database another_db
   transaction: false
```

## Ability to send queries to other databases in pg_yregress

Syntax-wise this can be as simple as:

```yaml
- query: select omni_ext_test.alloc_shmem_database_local()
   database: another_db
```

We would need to connect to manage another connection. This may not be too bad, but we need to think about connection reuse (may be not in the initial draft, though?)

_Originally posted by @yrashk in https://github.com/omnigres/omnigres/pull/278#discussion_r1383375407_

**Bounty size: $100**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

---

<details>
<summary><b>Don't want to start from scratch?</b></summary>
<br>

A change for kickstarting the process:

```diff
diff --git a/pg_yregress/pg_yregress.c b/pg_yregress/pg_yregress.c
index 481666ac0..74e1c9f80 100644
--- a/pg_yregress/pg_yregress.c
+++ b/pg_yregress/pg_yregress.c
@@ -54,6 +54,21 @@ static bool populate_ytest_from_fy_node(struct fy_document *fyd, struct fy_node
       }
     }
 
+    {
+      // Is the test meant to be ran on a different database?
+      struct fy_node *database = fy_node_mapping_lookup_by_string(test, STRLIT("database"));
+      y_test->database.base = NULL;
+
+      if (database != NULL) {
+        if (!fy_node_is_scalar(database)) {
+          fprintf(stderr, "database should be a string, got: %s",
+                  fy_emit_node_to_string(database, FYECF_DEFAULT));
+          return false;
+        }
+        y_test->database.base = fy_node_get_scalar(database, &y_test->database.len);
+      }
+    }
+
     {
       // Is the test meant to be negative? As in "if test succeeds, it's a failure"
       struct fy_node *negative = fy_node_mapping_lookup_by_string(test, STRLIT("negative"));
[...]

PR #399, opened 2023-12-11
Problem: impossible to run tests in a different database

rmodpur:
In some obscure cases (like omni_ext), we want to be able to test against multiple databases.

Currently, we can't do create database because we're wrapping everything into a transaction (CREATE DATABASE cannot run inside a transaction block error), and we can't switch databases on the fly (pg_yregress doesn't have support for it).

Solution: add transaction and database properties to pg_yregress test spec

setting `database` to a database name will create a connection to the specified database and close it at the end of the test

setting `transaction` to false (default is true) will not execute a `BEGIN` statement before running the test queries

#351

### DIFF
--- a/pg_yregress/pg_yregress.c
+++ b/pg_yregress/pg_yregress.c
@@ -70,6 +70,41 @@ static bool populate_ytest_from_fy_node(struct fy_document *fyd, struct fy_node 
     }
 
     {
+      // Is the test meant to be ran on a different database?
+      struct fy_node *database = fy_node_mapping_lookup_by_string(test, STRLIT("database"));
+      y_test->database.base = NULL;
+
+      if (database != NULL) {
+        if (!fy_node_is_scalar(database)) {
+          fprintf(stderr, "database should be a string, got: %s",
+                  fy_emit_node_to_string(database, FYECF_DEFAULT));
+          return false;
+        }
+        if (!managed) {
+          fprintf(stderr, "Can't use `database` in unmanaged instances\n");
+          return false;
+        }
+        y_test->database.base = fy_node_get_scalar(database, &y_test->database.len);
+      }
+    }
+
+    {
+      // Don't execute begin statement if false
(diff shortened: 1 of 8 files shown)


pg_yregress/pg_yregress.c
```diff
@@ -69,6 +69,41 @@ static bool populate_ytest_from_fy_node(struct fy_document *fyd, struct fy_node
...
+      if (database != NULL) {
+        if (!fy_node_is_scalar(database)) {
+          fprintf(stderr, "database should be a string, got: %s",
+                  fy_emit_node_to_string(database, FYECF_DEFAULT));
+          return false;
+        }
+        if (!managed) {
+          fprintf(stderr, "Can't use `database` in unmanaged instances\n");
+          return false;
+        }
```
rmodpur, line 86: Opted not to support managed instance because it may not necessarily be available at localhost and connecting to a different db would require additional password

yrashk: Perhaps worth mentioning in the documentation?

yrashk (approved):
Looks good overall. It will be worthwhile to update the changelog and the JSON schema to expose new properties.

The only reservation (or, rather, a thought/suggestion) I have is whether we should consider a case when we want to keep sessions going for new databases as opposed to finalizing the connection on each such test? Not sure if it is worth the hassle of maintaining a list of connections.

pg_yregress/pg_yregress.c
rmodpur: The Caveats part of unmanged instance is updated to include `database` property, should be enough right ?

yrashk: I was thinking further about this. When we're given an unmanaged instance, we do have its host, but you are right that the credentials required may be different.

 I think it's safe to continue the way you suggest, and if there will be an interest in supporting this for unmanaged instances, we can enable this behavior and extend `database` key to support a object to configure more than just the database; or support URLs.

PR #401, opened 2023-12-17
override `h2o_fatal` for `omni_httpc`

rmodpur:


extensions/omni_httpc/migrate/3_http_execute.sql
```diff
@@ -33,4 +33,9 @@ $$ language sql;
 create function http_execute_with_options(options http_execute_options, variadic requests http_request[]) returns setof http_response
 as
 'MODULE_PATHNAME',
-'http_execute' language c;
\ No newline at end of file
+'http_execute' language c;
+
+create function http_fatal() returns void
```
rmodpur, line 38: added just for testing will remove it later, would be good to have reliable way to trigger `h2o_fatal` to use in tests

Issue #402, opened 2023-12-19
Problem: omni_vfs backend support is still limited


yrashk:
Currently, omni_vfs only has only one backend: `local_fs`. This is very limiting as this is not as practical on the server.

Proposed solution: implement a `table_fs` that will provide the same API for table-backed filesystems.

Different designs can be considered, but it is important to ensure the performance of finding/listing files under a directory (recursively) as well as direct lookup.

I've had a (not a correctly implemented) attempt before: https://github.com/yrashk/omnigres/commit/b159f5631d84946af574c3b4faa5100b1b9001f2. Feel free to consult with it.

---

**Bounty size: $250**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

rmodpur:
Interested

PR #403, opened 2023-12-19
Problem: omni_vfs backend support is still limited

rmodpur:
Currently, omni_vfs only has only one backend: `local_fs`.
This is very limiting as this is not as practical on the server.

Solution: implement a `table_fs` that will provide the same API for table-backed filesystems

path parameters accepted by the functions is converted to absolute canonical form before usage

#402


extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,111 @@
...
+(
+    id   integer primary key generated always as identity,
+    name text not null unique
+);
+
+create table table_fs_files
+(
+    id            bigint primary key generated always as identity,
+    filesystem_id integer                not null references table_fs_filesystems (id),
+    filename      text[]                 not null,
```
rmodpur, line 11: for `filename` column instead of storing components of path, how about just storing absolute canonicalized path otherwise it requires a lot of array spliting and joining. Prefix matching should be enough right which can be provided by a btree index

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,155 @@
...
+                     chunk_size int default null) returns bytea
+    stable
+    language sql
+as
+$$
+with
+    filepath(path) as (select omni_vfs.canonicalize_path(path, absolute => true))
+    select
+        case
+            when chunk_size is null then substr(data, file_offset::int)
```
rmodpur, line 138: postgres inbuilt `substr` expects offset parameter to be of type `int` so there is a potential concern of truncation of value due to downcasting from `bigint` to `int`

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,155 @@
...
+
+
+-- TODO: refactor this into [sparse] [reusable] chunks
+create table table_fs_file_data
+(
+    file bigint primary key references table_fs_files (id),
+    /*  external storage optimizes fetching parts of data
+        at the expense of increased storage due to disabled compression
+    */
+    data bytea  storage external not null
```
rmodpur, line 42: should there be a limit for data length ? perhaps something that can fit inside an `int` type because postgres builtin substring functions expects `int` type for offset parameter

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,155 @@
...
+(
+    id   integer primary key generated always as identity,
+    name text not null unique
+);
+
+create table table_fs_files
+(
+    id            bigint primary key generated always as identity,
+    filesystem_id integer                not null references table_fs_filesystems (id),
+    filename      text                 not null,
```
rmodpur, line 11: should this renamed to `filepath` ?

extensions/omni_vfs/local_fs.c
```diff
@@ -66,7 +66,7 @@ static SPIPlanPtr insert_fs = NULL;
 
 Datum local_fs(PG_FUNCTION_ARGS) {
   if (PG_ARGISNULL(0)) {
-    ereport(ERROR, errmsg("must must not be NULL"));
+    ereport(ERROR, errmsg("mount must not be NULL"));
```
yrashk, line 69: Great catch, thanks! 😆

yrashk (commented):
I think it's a **great start**. I've tried it out manually. Seems to work.

One thing I noticed that on larger data sets, it's not particular fast. Not sure how avoid that is. I've added a million of records and the following took 10 seconds:

```sql
 select * from omni_vfs.list(omni_vfs.table_fs('test'), '/') order by name asc
```

Adding `limit 5` to it doesn't change the performance significantly (~8.5 seconds).

I wonder how much of this is a natural limitation of the API shape we have. Listing all files directly is also a bit slow but still faster (~3 seconds):

```sql
 select * from omni_vfs.table_fs_files order by filename asc
```

I was able to reduce the 3s to 1s by adding the following index:

```sql
create index i1 on omni_vfs.table_fs_files (filename)
```

The original omni_vfs query didn't get faster, though.

Would appreciate your thoughts on this.

**Performance aside**, a few more tests would definitely be in order, like testing list_recursively and whatever else might come to mind.

A **separate** **concern** here (stemming from my WIP) is that `table_fs` tables are in the extension's schema, so they won't be backed up or replicated by default. This is, of course, avoidable (see `pg_extension_config_dump`), but raises a question:

What if we allowed to define `table_fs_files`-compatible tables (in fact, we can provide a simple helper function that does that) in whatever regular schema and then `table_fs` type would not be a `int4` pointer to an ID in a table, but a `regclass` (effectively pointing to a table name)? This will prevent the table from being in the extension schema and will allow for more natural partitioning.

The only challenge with the PL/pgSQL part here is that we'd need to create a select query that works with different table names. Short term, we can do `format` to produce the query and then execute it, but that being an ugly practice, we can add "identifier rewriting" functionality to `omni_sql` (outside of the scope for this particular PR / bounty, of course). Alternatively, these functions can be done in C (again, this can be an optimization down the road).

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,217 @@
...
+    from
+        omni_vfs.table_fs_files
+    where
+        filesystem_id = new.filesystem_id and
+        (
+            -- file can't be inside a file
+            new.filepath like (filepath || '/%')
+            or
+            -- dir can't be a file
+            filepath like (new.filepath || '/%')
```
rmodpur, line 37: both are necessary to ensure correctness, but it will become slower as the table grows because `new.filepath like (filepath || '/%')` always leads to a sequential scan

yrashk: can we enforce these rules on inserts/updates to the table?

rmodpur: It is enforced on inserts and updates (of some columns) only

yrashk: I apologize, I misread the name of the function above. I see this is the trigger itself. So the problem identified here is on the trigger side making inserts/updates, slow, correct?

That is, it is separate from the performance of listing files. Correct?

rmodpur: Yes, this is separate from listing performance

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,217 @@
...
+                    when split_part(substr(filepath, length(search.path)+1), '/', 3) = '' then
+                        row(split_part(substr(filepath, length(search.path)+1), '/', 2), kind)::omni_vfs_types_v1.file
+                    -- more than one '/', eg. /dir/file
+                    else row(split_part(substr(filepath, length(search.path)+1), '/', 2), 'dir')::omni_vfs_types_v1.file
+                end
+        )
+    end
+from omni_vfs.table_fs_files, search
+where filesystem_id = fs.id and
+(filepath like (search.path || '/%') or
```
rmodpur, line 155: using `pg_trgm` to create index on `filepath` should help with `like` operator but listing of paths such `/` will most likely lead to seq scan because we store entire filepath in each entry, leading to every entry matching(postgres planner may avoid using indexes in such cases)  just to produce few directories and files that are present in root directory

yrashk: It doesn't really have to be a `like` operator but just substring, really. Does this help us?

Aside, I wonder if (at least theoretically, as this is beyond the scope of this PR) we can implement a custom index.

rmodpur: Yes, it can be converted to use substring but the issue still remains that postgres planner sees too many entries are matching for certain entries so it will opt for seq scan.

It would have been ok if most of the entries in the root directory are different but filesystems usually have some small number of directories in root. 

For eg. if the following files are in the table `/dir/file1`, `/dir/file2`, `/dir/file3`, we end up processing all the three entries just to produce `dir` when listing `/`

yrashk:
Assuming you're sure this code is fine functionally (think of ways to achieve incorrect behaviour, write more tests for edge-cases), can you please do the following?

* Prepare a polished commit
* Extract precise definitions (with query plans, if necessary) of performance problems with the current approach and create a separate issue. Can reference this PR in it to keep the connection. We can discuss solving them there.

Have you had a chance to think about my proposal of "table_fs_files-compatible tables" from above? Quoting it below for convenience.

>A separate concern here (stemming from my WIP) is that table_fs tables are in the extension's schema, so they won't be backed up or replicated by default. This is, of course, avoidable (see pg_extension_config_dump), but raises a question:
>
> What if we allowed to define table_fs_files-compatible tables (in fact, we can provide a simple helper function that does that) in whatever regular schema and then table_fs type would not be a int4 pointer to an ID in a table, but a regclass (effectively pointing to a table name)? This will prevent the table from being in the extension schema and will allow for more natural partitioning.
>
>The only challenge with the PL/pgSQL part here is that we'd need to create a select query that works with different table names. Short term, we can do format to produce the query and then execute it, but that being an ugly practice, we can add "identifier rewriting" functionality to omni_sql (outside of the scope for this particular PR / bounty, of course). Alternatively, these functions can be done in C (again, this can be an optimization down the road).

rmodpur:
> Have you had a chance to think about my proposal of "table_fs_files-compatible tables" from above?

Not yet, will work on it next

rmodpur:
Test failures are interesting, tests are passing in my computer. Looks like string sort order is different between my computer and CI runners

extensions/omni_vfs/migrate/4_table_fs.sql
```diff
@@ -0,0 +1,217 @@
...
+                        row(split_part(substr(filepath, length(search.path)+1), '/', 2), kind)::omni_vfs_types_v1.file
+                    -- more than one '/', eg. /dir/file
+                    else row(split_part(substr(filepath, length(search.path)+1), '/', 2), 'dir')::omni_vfs_types_v1.file
+                end
+        )
+    end
+from omni_vfs.table_fs_files, search
+where filesystem_id = fs.id and
+(starts_with(filepath, search.path || '/') or
+filepath = search.path)
```
yrashk, line 156: Could the tests be failing because `list` here does not establish any ordering?

yrashk:
Superseded by #429

PR #404, opened 2023-12-19
Problem: omni_txn's scope is limited

### LINKED ISSUE #398: Problem: omni_txn's scope is limited
Presently, omni_txn only provides functionality to declare transaction-scoped typed variables. It provides no functionality actually to do something with and around transactions. 

It's both limiting and confusing.

Proposed solution: extract variable functionality into a different extension (naming TBD) and extend it with session-scoped and durable variables. 

Keep omni_txn to do something useful with transactions. Here are just some ideas:

* Explicit transactional pre- and post-conditions?
* Any kind of easily processable transaction analytics?

---

### Notes on expanded variable scoping

It's fairly trivial to make variables **session**-scoped. In that case, we need not enforce transactional boundaries like we do now. **Durable** variables are a little bit more interesting. We could store them as their binary encoding (if available; textual encoding otherwise) in a table alongside the type name.

It would be important to design a good API around these different scopes that would not be confusing.

yrashk:
Presently, omni_txn only provides functionality to declare transaction-scoped typed variables. It provides no functionality actually to do something with and around transactions.

It's both limiting and confusing.

Solution: extract this functionality into omni_var extension

omni_txn.get_variable and omni_txn.set_variable are deprecated now.

Closes #398


yrashk:
Pre-merge plan:

- [ ] should `set` return the previously set value?
- [ ] session scoping
- [ ] durable scoping

yrashk:
> * should `set` return the previously set value?

I think the original motivation behind `set` returning the new value was so that the new value could be pipelined further up.

But whether this is a good idea or not remains to be seen.

Can we make this configurable through a parameter?

```sql
set('var', return_previous => true)
```

`return_previous` is a horrible name, but conceptually, is this better?

extensions/omni_var/docs/variables.md
```diff
@@ -0,0 +1,56 @@
...
+This is most useful to maintain information across multiple queries in the
+transaction, particularly with RLS (Row Level Security) policies in mind.
+
+## Setting a variable
+
+Within a transaction's context, one can set a named variable with its type
+specified through the type of the value:
+
+```postgresql
+select omni_var.set_variable('my_variable', true)
```
viggy28, line 15: Isn't this just set()?

viggy28:
> But whether this is a good idea or not remains to be seen.

To me,`set()` is something like a put operation and I don't expect it to return the previous value. 

Since we are planning to add support for session and durable scoping, wonder how's that going to impact the current syntax of `set()` and `get()`.

yrashk:
> > But whether this is a good idea or not remains to be seen.
> 
> To me,`set()` is something like a put operation, and I don't expect it to return the previous value.

`set()` does, but the one returning the previous value would be rather called `replace()`, I suppose. The reason why I am thinking about this is that basically if we _do_ want to know the previous value, then we need to lookup the hash table twice, but if we return the previous value, then it is only one lookup.
 
> Since we are planning to add support for session and durable scoping, wonder how's that going to impact the current syntax of `set()` and `get()`.

I was thinking alone the lines of:

```sql
omni_var.set('var', null::int4, session => true)
omni_var.set('var', null::int4, durable => true)
```

But perhaps `get` would always get the "closest" scope: transaction->session->durable.

Thoughts?

viggy28:
> set() does, but the one returning the previous value would be rather called replace(), I suppose. The reason why I am thinking about this is that basically if we do want to know the previous value, then we need to lookup the hash table twice, but if we return the previous value, then it is only one lookup.

or can we combine `set()` and `replace()` with one operation `upsert()` which inserts if there is no existing value and updates if there is and returns the old value? If the old value is null, then there is no existing value.

The syntax looks good for defining the scope of the variable. However, I am not sure about the assumption for `get()` returning the closest scope. It might catch people by surprise.

yrashk:
> > set() does, but the one returning the previous value would be rather called replace(), I suppose. The reason why I am thinking about this is that basically if we do want to know the previous value, then we need to lookup the hash table twice, but if we return the previous value, then it is only one lookup.
> 
> or can we combine `set()` and `replace()` with one operation `upsert()` which inserts if there is no existing value and updates if there is and returns the old value? If the old value is null, then there is no existing value.

Would that mean `set()` will fail if not present? 

What if we let `set()` behave as it does but add `set(name, value, update => true)` which will require that the previous value is set and returns the previous value; and raises an error otherwise? Or `update()` to the same effect.

> 
> The syntax looks good for defining the scope of the variable. However, I am not sure about the assumption for `get()` returning the closest scope. It might catch people by surprise.

What if allow specifying the scope explicitly and implicitly go for the nearest available?

extensions/omni_var/docs/variables.md
yrashk: Yes

extensions/omni_txn/docs/variables.md
```diff
@@ -1,5 +1,10 @@
 # Transaction Variables
 
+!!! warning "This functionality has been moved"
```
viggy28, line 3: I am happy with even removing `omni_txn` completely in this PR. Not sure we need to worry about backward compatibility at this point.

extensions/omni_var/omni_var.c
```diff
@@ -0,0 +1,231 @@
...
+  VariableValue variable_value;
+} Variable;
+
+static int nvars = 0;
+
+void _PG_init() {
+  DefineCustomIntVariable(
+      "omni_var.estimated_initial_txn_variables_count", "Estimated number of transaction variables",
+      "Estimated number of transaction variables, changes apply to the next transaction", &nvars,
+      1024, 0, 65535, PGC_USERSET, 0, NULL, NULL, NULL);
```
viggy28, line 45: (Given transactionId is 32 bit integer) Is 65535 (2^16) the maximum number of variables that one can define within a transaction?

extensions/omni_var/omni_var.c
```diff
@@ -0,0 +1,231 @@
...
+    }
+  default:
+    // nothing to do
+    return;
+  }
+}
+
+Datum set(PG_FUNCTION_ARGS) {
+  if (PG_ARGISNULL(0)) {
+    ereport(ERROR, errmsg("variable name must not be a null"));
```
viggy28, line 133: should we also make sure that it is of type `text` only or does it support other types too?

viggy28 (approved):

extensions/omni_txn/docs/variables.md
yrashk: We have users using it right now, don't want to break it right away

extensions/omni_var/omni_var.c
yrashk: This is not a limit, but estimation and transaction IDs have nothing to do with the estimation or limit.

extensions/omni_var/omni_var.c
yrashk: See the documentation and thr Variable type. It's intentionally and expressly typed.

PR #414, opened 2023-12-23
Problem: intermittent CI problem with http2_reproxy

yrashk:
This is the test: https://github.com/omnigres/omnigres/blob/master/extensions/omni_httpd/tests/http2_reproxy.yml

It often fails due to a timeout in CI, and we have to re-run it.

The behaviour it is testing is as follows:

Considering that we have only two workers (https://github.com/omnigres/omnigres/blob/master/extensions/omni_httpd/tests/http2_reproxy.yml#L6)

If we send two HTTP/2 requests over the same connection, instead of blocking on the second request (since the Postgres worker is busy), omni_httpd will reproxy the request to the other available worker.

It is tested by making the backend sleep for 2 seconds. The test ensures that it takes no longer than 2.5 seconds total responses return different pids for their backends.  I wondered if this can be tested better (preferably without timeouts involved) or if the implementation is deficient. This feature has undergone manual testing before using a command line HTTP client to observe the behaviour.

Solution: rewrite the test to not be time-based

Instead use locks and sequences to check whether the requests are done in parallel.

Closes #100

### DIFF
--- a/extensions/omni_httpd/tests/http2_reproxy.yml
+++ b/extensions/omni_httpd/tests/http2_reproxy.yml
@@ -8,6 +8,55 @@ instance:
   - set session omni_httpd.no_init = true
   - create extension omni_httpd cascade
   - create extension omni_httpc cascade
+  - create sequence test_sequence;
+  - |
+    create function handler(request omni_httpd.http_request) returns omni_httpd.http_outcome
+        language plpgsql
+    as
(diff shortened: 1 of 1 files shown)


yrashk:
I cannot reproduce frequent CI errors independently (the infamous "works on my machine"). Investigating.

PR #417, opened 2023-12-25
Problem: locale setting of database is dependent on environment

### LINKED ISSUE #416: Problem: `locale` setting of database is dependent on environment
[omni_vfs table_fs](https://github.com/omnigres/omnigres/pull/403) tests failed due to inconsistent sort ordering of text between my machine and CI runners.

The reason turned out to be different `LC_COLLATE`(responsible for String sort order) was different `en_US.UTF-8` (on my machine) and `C.UTF-8` (on CI runners) because

from [postgres documentation](https://www.postgresql.org/docs/15/locale.html)

> Locale support is automatically initialized when a database cluster is created using initdb. initdb will initialize the database cluster with the locale setting of its execution environment by default
When we speak of inheriting the locale from the execution environment, this means the following on most operating systems: For a given locale category, say the collation, the following environment variables are consulted in this order until one is found to be set: LC_ALL, LC_COLLATE (or the variable corresponding to the respective category), LANG. If none of these environment variables are set then the locale defaults to C

We are not explicitly setting any locale in our 
[initdb command](https://github.com/omnigres/omnigres/blob/master/cmake/PostgreSQLExtension.cmake#L485)

from [postgres documentation](https://www.postgresql.org/docs/15/locale.html)

> If you want to use a different locale (or you are not sure which locale your system is set to), you can instruct initdb exactly which locale to use by specifying the --locale option

rmodpur:
omni_vfs table_fs tests failed due to inconsistent sort ordering of text between my machine and CI runners.

The reason turned out to be different LC_COLLATE(responsible for String sort order) was different en_US.UTF-8 (on my machine) and C.UTF-8 (on CI runners) because initdb will initialize the database cluster with the locale setting of its execution environment by default.

Solution: explicitly set encoding and locale for initdb to avoid any surprises

Closes #416

### DIFF
--- a/pg_yregress/schema.json
+++ b/pg_yregress/schema.json
@@ -55,6 +55,14 @@
         "hba": {
           "type": "string",
           "description": "pg_hba.conf config"
+        },
+        "encoding": {
+          "type": "string",
+          "description": "default encoding for database"
+        },
+        "locale": {
+          "type": "string",
+          "description": "default locale for database"
         }
       }
     },
--- a/pg_yregress/test.yml
+++ b/pg_yregress/test.yml
@@ -26,6 +26,9 @@ instances:
     hba: |
       local all all trust
       host all all all trust
+  encoding_locale:
+    encoding: SQL_ASCII
+    locale: en_US.UTF-8
 
 tests:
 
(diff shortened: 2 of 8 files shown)


pg_yregress/test.yml
```diff
@@ -26,6 +26,9 @@ instances:
     hba: |
       local all all trust
       host all all all trust
+  encoding_locale:
+    encoding: EUC_KR
+    locale: ko_KR
```
rmodpur, line 31: need some encoding and locale which will be present in CI runners as well

yrashk: @suryaoruganti I wonder if you have thoughts on locales availability? Or is this how GH worker images are, and they replicated as closely as possible?

pg_yregress/schema.json
```diff
@@ -55,6 +55,14 @@
...
           "type": "string",
           "description": "pg_hba.conf config"
+        },
+        "encoding": {
+          "type": "string",
+          "description": "default encoding for database"
+        },
+        "locale": {
+          "type": "string",
+          "description": "default locale for database"
```
yrashk, line 65: Perhaps it is worth specifying defaults here (`UTF8/C`), what do you think?

yrashk (approved):

pg_yregress/schema.json
rmodpur: I thought about it too but didn't mention because if someone cares about encoding and locale they should specify it anyways even if they match the defaults

pg_yregress/test.yml
suryaoruganti: Our runners are built from the same source as GitHub upstream. However, there could be some additional configs that are not a part of the open source that GitHub could be doing. I'll check this.

Thanks for bringing this to my notice.

PR #418, opened 2023-12-25
Problem: configuring psql target dbs

yrashk:
Currently it's impossible to configure the databases `psql_` targets create. However, this is very limiting because sometimes we want to use them with a particular configuration.

Solution: use postgresql.conf and pg_hba.confi if provided

### DIFF
--- a/cmake/PostgreSQLExtension.cmake
+++ b/cmake/PostgreSQLExtension.cmake
@@ -484,7 +484,26 @@ export EXTENSION_SOURCE_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}\"
 rm -rf \"${CMAKE_CURRENT_BINARY_DIR}/data/${NAME}\"
 ${INITDB} -D \"${CMAKE_CURRENT_BINARY_DIR}/data/${NAME}\" --no-clean --no-sync --locale=C --encoding=UTF8
 export SOCKDIR=$(mktemp -d)
-echo host all all all trust >>  \"${CMAKE_CURRENT_BINARY_DIR}/data/${NAME}/pg_hba.conf\"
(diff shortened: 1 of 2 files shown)


cmake/PostgreSQLExtension.cmake
```diff
@@ -484,7 +484,26 @@ export EXTENSION_SOURCE_DIR=\"${CMAKE_CURRENT_SOURCE_DIR}\"
...
-echo host all all all trust >>  \"${CMAKE_CURRENT_BINARY_DIR}/data/${NAME}/pg_hba.conf\"
+if [ -z \"$POSTGRESQLCONF\" ]; then
+  if [ -f \"${CMAKE_CURRENT_SOURCE_DIR}/postgresql.conf\" ]; then
+    export POSTGRESQLCONF=\"${CMAKE_CURRENT_SOURCE_DIR}/postgresql.conf\"
+  fi
+fi
+echo \"Using postgresql.conf: $POSTGRESQLCONF\"
+if [ -z \"$PGHBACONF\" ]; then
+ if [ -f \"${CMAKE_CURRENT_SOURCE_DIR}/pg_hba.conf\" ]; then
+    export PGHBACONF=\"${CMAKE_CURRENT_SOURCE_DIR}/postgresql.conf\"
```
viggy28, line 495: shouldn't this be `pg_hba.conf`?

viggy28 (changes requested):

cmake/PostgreSQLExtension.cmake
yrashk: It should be

viggy28 (approved):

PR #428, opened 2023-12-30
Problem: Multi DB tests still rely on pg_regress

idoqo:
Problem: Multi DB tests still rely on pg_regress

Solution: Rewrite existing tests using pg_yregress' support for separate databases. Additionally, this also rewrites libpgaug's tests and should complete our migration.


extensions/omni_ext/test/tests/omni_ext_test/tests.yml
```diff
@@ -42,9 +53,34 @@ tests:
     - query: select omni_ext_test.wait_for_table('local_worker_started')
       results:
       - wait_for_table: true
-  
+    - query: select n.nspname, c.relname from pg_catalog.pg_class c left join pg_catalog.pg_namespace  n on n.oid=c.relnamespace where c.relname='local_worker_started'
```
idoqo, line 56: @yrashk turned out we can't exactly re-implement describe (`\d`) with only SQL, I'm not sure what its role was in the pg regress tests, so I've only added these queries here to check existence of the table.

yrashk: Can you please elaborate on what exactly can't be implemented when it comes to `\d`?

yrashk: As for the role of those tests, do you mean this part?

```
\d local_worker_started;
      Table "public.local_worker_started"
 Column | Type | Collation | Nullable | Default 
--------+------+-----------+----------+---------
```

If yes, then yeah, we're just checking for the presence.

idoqo: > Can you please elaborate on what exactly can't be implemented when it comes to \d?

One immediate example was how \d supports describing other objects aside from tables (views, sequences, etc). Since we are only using it to check for presence though, I think the current query works and it's no longer a problem.

yrashk: Yeah, I think that use of `\d` was just a "lazy" version of `\dt`

yrashk:
Great work! 👍

Issue #433, opened 2024-01-03
Problem: tablefs backend for omni_vfs discovery


yrashk:
`tablefs` can only be discovered by reading the source code of `omni_vfs`, which is ultimately not great for DX.

Proposed solution: briefly document it

cc @rmodpur

yrashk:
Resolved in #439

PR #439, opened 2024-01-04
Problem: table_fs backend for omni_vfs discovery

rmodpur:
table_fs can only be discovered by reading the source code of omni_vfs, which is ultimately not great for DX.

Solution: document its usage

#433

### DIFF
--- a/extensions/omni_vfs/docs/reference.md
+++ b/extensions/omni_vfs/docs/reference.md
@@ -152,6 +152,64 @@ This backend can be created by invoking `omni_vfs.local_fs(mount)`, where `mount
     under which conditions mounting of new directories is possible, and which mounts can be accessed under given
     conditions.
 
+### `omni_vfs.table_fs` (table backed file system)
+
+This backend can be created by invoking `omni_vfs.table_fs('fs')`, where `fs` is the name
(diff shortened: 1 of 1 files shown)


extensions/omni_vfs/docs/reference.md
```diff
@@ -152,6 +152,64 @@ This backend can be created by invoking `omni_vfs.local_fs(mount)`, where `mount
...
+
+* The following query creates an entry for sample.txt file in `omni_vfs.table_fs_files` in `fs` table_fs filesystem. There is no need to create parent directories, they will be
+created automatically if it doesn't exist.
+```postgresql
+insert into omni_vfs.table_fs_files (filesystem_id, filename, kind)
+values 
+((omni_vfs.table_fs('fs')).id, '/dir/sample.txt', 'file');
+```
+
+* To associate data with the file created above run the following query. It creates an entry in `omni_vfs.table_fs_file_data`. A utility function named `omni_vfs.table_fs_file_id` is provided to obtain file_id given a table_fs filesystem and filename. Only a single data entry can be associated with a given file.
```
yrashk, line 177: This illuminates the fact that we need write side of VFS API, too :)

rmodpur: Yes and probably removal of files as well

PR #440, opened 2024-01-04
Problem: migrating extensions with code in stored procedures

yrashk:
To enable proper migration between arbitrary revisions, we'd need to make a separate numbered migration for every stored procedure whenever that procedure's code changes.

This is really inconvenient.

Solution: extract actual stored procedures code into source files and refer to them from migrations.

This way, when we assemble migration in any given revision, we can get the right code for it. The most important scenario of all:

Imagine we have an extension A at revision R0 with some structual migrations and one stored procedure. New revision of this extension (R1) changes the code of that stored procedure and nothing else.

The script to create extension at this revision will be correct immediately as we'll pull the correct code from a separate source file.

What's the upgrade path here between R0 and R1? It's a bit involved, but please stay with me:

We first provision a database with R0 and record source code in all functions created by the extension (avoiding C-language ones). Then, we provision R1 and do the same. For every function that exists in both but has a different source code, we need to replace those function definitions with the new ones. `pg_get_functiondef` covers us here.


rmodpur:
so `create or replace function` is not suitable ?

yrashk:
The purpose of this change is to avoid multiple incremental migrations changing the same function source code

```
100_replace_myfunc.sql
101_replace_myfunc.sql
...
```

Instead, put the actual source (SQL, PL/pgSQL, Python, etc.) into separate files and include them from the migration template.

Issue #479, opened 2024-02-01
Problem: pg_rewind fails in presence of omni_httpd extension files in postgres data directory


rmodpur:
It is impossible to restore from pg_rewind in the presence of omni_httpd\\\* directories

![image](https://uploads.linear.app/eb9d5cd3-3c40-4a58-a699-232cf39fe69b/d195447b-1938-42b3-8e9f-6a272fe77e96/3e2f0d09-dc19-436a-bffc-e40a4b4b8198?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2ViOWQ1Y2QzLTNjNDAtNGE1OC1hNjk5LTIzMmNmMzlmZTY5Yi9kMTk1NDQ3Yi0xOTM4LTQyYjMtOGU5Zi02YTI3MmZlNzdlOTYvM2UyZjBkMDktZGMxOS00MzZhLWJmZmMtZTQwYTRiNGI4MTk4IiwiaWF0IjoxNzA3MTcwMDU4LCJleHAiOjMzMjc3NzMwMDU4fQ.Jd6LuwmRJ8iAbIsKETJmVEYu6jyNLjUh9calRk7Qc0c)

PR #480, opened 2024-02-01
Problem: pg_rewind fails in presence of omni_httpd extension files in postgres data directory

rmodpur:
Solution: use temp directory to store omni_httpd unix socket file

#479

### DIFF
--- a/extensions/omni_httpd/master_worker.c
+++ b/extensions/omni_httpd/master_worker.c
@@ -196,9 +196,14 @@ static void sigusr1_handler(int signo) {
  */
 void master_worker(Datum db_oid) {
   IsOmniHttpdWorker = true;
-  char socket_path_template[] = "omni_httpdXXXXXX";
-  char *tmpname = mkdtemp(socket_path_template);
-  socket_path = psprintf("%s/socket.%d", tmpname, getpid());
+  char *tmp_path = mkdtemp(
+      psprintf("%s/%s", GetConfigOption("omni_httpd.temp_dir", false, false), "omni_httpdXXXXXX"));
+  if (tmp_path == NULL) {
+    ereport(ERROR, errmsg("'%s' isn't a directory, please set omni_httpd.temp_dir to an existing "
+                          "temporary directory",
+                          GetConfigOption("omni_httpd.temp_dir", false, false)));
+  }
+  socket_path = psprintf("%s/socket.%d", tmp_path, getpid());
 
 #if PG_MAJORVERSION_NUM >= 13
   pqsignal(SIGHUP, SignalHandlerForConfigReload);
--- a/extensions/omni_httpd/tests/create_drop.yml
+++ b/extensions/omni_httpd/tests/create_drop.yml
@@ -10,5 +10,42 @@ tests:
   tests:
   - query: create extension omni_httpd cascade
     commit: true
+
+    # https://github.com/omnigres/omnigres/issues/479
+  - query: select count(*)
+           from pg_ls_dir((select setting from pg_settings where name = 'data_directory'))
+           where pg_ls_dir like 'omni_httpd%'
+    results:
+      - count: 0
   - query: drop extension omni_httpd
     commit: true
+
+- name: temp_dir doesn't exist
+  transaction: false
+  query: alter system set omni_httpd.temp_dir = '/does-not-exist'
+  error:
+    severity: ERROR
+    message: "'/does-not-exist' temp directory does not exist."
+
+- name: temp_dir inside data dir
+  transaction: false
+  #  unable to execute alter system with dynamic value in SQL but this has been tested manually
+  todo: true
+  query: |
+    do
+    $$
+        declare
+            data_dir text;
+        begin
+            select setting
+            into data_dir
+            from pg_settings
+            where name = 'data_directory';
+            
+            execute format('alter system set omni_httpd.temp_dir = ''%s''', data_dir);
+        end;
+    
+    $$ language plpgsql;
+  error:
+    severity: ERROR
+    message: temp directory location should not be inside the data directory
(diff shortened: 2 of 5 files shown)


extensions/omni_httpd/master_worker.c
```diff
@@ -196,9 +196,19 @@ static void sigusr1_handler(int signo) {
  */
 void master_worker(Datum db_oid) {
   IsOmniHttpdWorker = true;
-  char socket_path_template[] = "omni_httpdXXXXXX";
-  char *tmpname = mkdtemp(socket_path_template);
-  socket_path = psprintf("%s/socket.%d", tmpname, getpid());
+  char *tmpdir = getenv("TMPDIR");
+  if (tmpdir == NULL || strcmp(tmpdir, "") == 0) {
+    // fallback to /tmp
```
yrashk, line 201: How about we try to see if there is a configuration variable... maybe `omni_httpd.temp_dir` that is set before falling back to /tmp?

extensions/omni_httpd/tests/create_drop.yml
```diff
@@ -3,12 +3,21 @@ instance:
...
 tests:
 
 - name: shuts down after drop
   tests:
   - query: create extension omni_httpd cascade
     commit: true
+
+    # https://github.com/omnigres/omnigres/issues/479
+  - query: select count(*)
+           from omni_vfs.list(omni_vfs.local_fs((select setting from pg_settings where name = 'data_directory')), '/')
```
yrashk, line 18: I think here in the text we can get away with `pg_ls_dir` instead of omni_vfs?

yrashk (commented):
I found that there's [TempTablespacePath](https://doxygen.postgresql.org/fd_8c.html#aff120e0e4e406c3836d8a0476bbae2af)

Worth a quick research to see if the dedicated temp path in the tablespace would be tolerated by pg_rewind?

rmodpur:
> I found that there's [TempTablespacePath](https://doxygen.postgresql.org/fd_8c.html#aff120e0e4e406c3836d8a0476bbae2af)
> 
> Worth a quick research to see if the dedicated temp path in the tablespace would be tolerated by pg_rewind?

I considered using it but found out it is used by postgres to spill data that can't fit in memory for eg. sorting

extensions/omni_httpd/master_worker.c
rmodpur: We can do that, I just followed the [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap10.html)

> The /tmp directory is retained in POSIX.1-2017 to accommodate historical applications that assume its availability. Implementations are encouraged to provide suitable directory names in the environment variable TMPDIR and applications are encouraged to use the contents of TMPDIR for creating temporary files.

extensions/omni_httpd/tests/create_drop.yml
rmodpur: Wasn't aware of it, have replaced omni_vfs with pg_ls_dir, thanks for pointing out

extensions/omni_httpd/master_worker.c
rmodpur: even postgres test suite does the same thing
https://doxygen.postgresql.org/pg__regress_8c.html#ad9e2fb61733462b15438cffb898f4b56

rmodpur: ok, let's just use `omni_httpd.temp_dir` with it's default value set to `/tmp`

rmodpur: i'm using a custom guc now

extensions/omni_httpd/tests/create_drop.yml
```diff
@@ -10,5 +10,40 @@ tests:
...
+    $$
+        declare
+            data_dir text;
+        begin
+            select setting
+            into data_dir
+            from pg_settings
+            where name = 'data_directory';
+            
+            execute format('alter system set omni_httpd.temp_dir = ''%s''', data_dir);
```
rmodpur, line 43: this test fails with `ALTER SYSTEM cannot be executed from a function`
@yrashk any ideas on how to execute 'ALTER SYSTEM` with dynamic value ?

yrashk: may be call `set_config(...)` instead?

rmodpur: doesn't work
`ERROR:  parameter "omni_httpd.temp_dir" cannot be changed now`

extensions/omni_httpd/master_worker.c
```diff
@@ -196,9 +196,14 @@ static void sigusr1_handler(int signo) {
  */
 void master_worker(Datum db_oid) {
   IsOmniHttpdWorker = true;
-  char socket_path_template[] = "omni_httpdXXXXXX";
-  char *tmpname = mkdtemp(socket_path_template);
-  socket_path = psprintf("%s/socket.%d", tmpname, getpid());
+  char *tmp_path = mkdtemp(
+      psprintf("%s/%s", GetConfigOption("omni_httpd.temp_dir", false, false), "omni_httpdXXXXXX"));
+  if (tmp_path == NULL) {
+    ereport(ERROR, errmsg("'%s' isn't a directory, please set omni_httpd.temp_dir to an existing "
```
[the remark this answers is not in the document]
rmodpur: @yrashk any comments on this behavior ?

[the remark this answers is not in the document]
yrashk: Well, since we have a check, it should be fine, right?

extensions/omni_httpd/tests/create_drop.yml
rmodpur: it has been tested manually, marked it as todo

PR #481, opened 2024-02-05
Problem: omni_httpd unlocks listener table twice

rmodpur:
When started with no default config
(set session omni_httpd.no_init = true before create extension omni_httpd cascade) I observed this warning:

`WARNING:  you don't own a lock of type ExclusiveLock`

Solution: unlock listener table only if it's locked

### DIFF
--- a/extensions/omni_httpd/master_worker.c
+++ b/extensions/omni_httpd/master_worker.c
@@ -260,6 +260,7 @@ void master_worker(Datum db_oid) {
 
     Oid nspoid = get_namespace_oid("omni_httpd", false);
     Oid listenersoid = get_relname_relid("listeners", nspoid);
+    bool listenersoid_locked = false;
 
     // Get listeners
     while (worker_reload) {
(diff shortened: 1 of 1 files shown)


yrashk:
I wonder if we can make this deterministic? We know at which points we exit the `while` loop, does the holding of lock become non-deterministic there to the point we need to have a flag?

rmodpur:
> does the holding of lock become non-deterministic there to the point we need to have a flag?

It certainly can be done :point_down:  without introducing a new flag but I prioritized readability of code
```diff
diff --git a/extensions/omni_httpd/master_worker.c b/extensions/omni_httpd/master_worker.c
index b998845..d46414f 100644
--- a/extensions/omni_httpd/master_worker.c
+++ b/extensions/omni_httpd/master_worker.c
@@ -473,7 +473,7 @@ void master_worker(Datum db_oid) {
     }
 
     // Unlock the table at last
-    if (listenersoid_locked) {
+    if (port_ready) {
       UnlockRelationOid(listenersoid, ExclusiveLock);
       listenersoid_locked = false;
     }
```

yrashk:
Getting rid of an extra flag is better, but I meant, could we figure out our loop exit in such a way that we don't need to check any flag? Less `if`s. It's not a highly critical part; just bothers me a bit that we just did something (lock/unlock), and now we have to consult with some flag.

A bigger problem with this code [not the PR itself] (as I noticed before and now again) is that it actually doesn't try to wait until port is ready, and starts the http workers. Ideally, though, shouldn't we wait until we have anything to start for? If there are no listeners to listen on, it's going to be a waste.

PR #486, opened 2024-02-14
Problem: omnigres extensions are released at each commit

rmodpur:
Omnigres extensions are versioned using commit
SHAs and released at each commit. This release
cadence is too frequent and it's not obvious to
compare commit versions without looking at git
history. Another drawback is all the extensions
share the same version.

Solution: Use semver versions of the form x.y.z
for omnigres extensions

With this each extension will have it's own independent
version. Only the extensions with new release will be packaged
and uploaded for distribution. New release of an
extension can be cut by creating/updating an entry
of the form \<ext-name>=x.y.z in 'versions.txt' file.


rmodpur:
1. Version of an extension is passed through cmake function `add_postgresql_extension`
2. `check_extension_artifacts` python script is for upholding certain invariants related to artifacts.txt enforced at the end of cmake configure step
3. extension release action downloads an index.json file from s3 bucket. It contains mapping of extension versions to commit shas. Version of an extension is parsed through artifacts.txt and it's upgrade script is only created and uploaded if the version is not present in index.json. For eg. if `omni_httpd` versions 1.1.0 and 1.2.0 have already been released and a new version 1.1.1 is created then the following upgrade scripts will be created: `omni_httpd--1.1.0--1.1.1` and `omni_httpd--1.1.1--1.2.0` i.e between the new version and any existing version

rmodpur:
`generate-upgrades.sh` - generates upgrade SQL scripts for an extension's new release with it's existing releases
`tools/check_extension_artifacts.py` - checks for cycles in extension dependency graph
`tools/package_extension_with_dependencies.py` - Downloads an extension dependencies and packages it along with the extension

PR #501, opened 2024-02-28
Problem: CMake cannot pass compiler optimization flags

JoshInnis:
When using Cmake to compile Omnigres, there is no ability to pass in compiler optimization flags. This is an issue because it can greatly slow down performance for release versions on Omnigres.

Solution: pass CMake's build-type-specific cflags to Postgres

### DIFF
--- a/cmake/FindPostgreSQL.cmake
+++ b/cmake/FindPostgreSQL.cmake
@@ -135,9 +135,9 @@ if(NOT DEFINED PG_CONFIG)
         endif()
 
         if(BUILD_TYPE STREQUAL "RelWithDebInfo")
-            set(ENV{CFLAGS} "$ENV{CFLAGS} -O2")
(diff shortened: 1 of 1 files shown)


yrashk:
Typo: ans->and

cmake/FindPostgreSQL.cmake
```diff
@@ -135,6 +135,12 @@ if(NOT DEFINED PG_CONFIG)
             string(APPEND extra_configure_args " --enable-debug --enable-cassert")
         endif()
 
+        if(BUILD_TYPE STREQUAL "RelWithDebInfo")
+            set(ENV{CFLAGS} "$ENV{CFLAGS} -O2")
+        elseif(BUILD_TYPE STREQUAL "Release")
+            set(ENV{CFLAGS} "$ENV{CFLAGS} -O2")
```
yrashk, line 141: I suspect the correct variables to use here would be 
`CMAKE_C_FLAGS_RELEASE` and `CMAKE_C_FLAGS_RELWITHDEBINFO` instead of -O2

PR #504, opened 2024-02-29
Problem: only package and release extension versions changed in a PR

rmodpur:
It is not trivial to do it by storing extension versions in CMakeLists.txt.

Solution: Store current extension versions in 'versions.txt' file. Each line in it will be in the form of \<ext-name>=\<ext-ver>. Any extension not specified in it will be assigned a special 'unreleased' version. This file will be used to compute the extension versions changed in a PR.

Required for extension release PR (#486)


cmake/Version.cmake
```diff
@@ -22,9 +22,10 @@ macro(get_version NAME VERSION_VAR)
     elseif(DEFINED OMNIGRES_VERSION)
         set("${VERSION_VAR}" "${OMNIGRES_VERSION}")
     else()
-        execute_process(COMMAND git describe  --tags --always --match "^v"
-                WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+        execute_process(COMMAND bash -c "./get-extension-version.sh versions.txt ${_name}"
```
yrashk, line 25: Is there anything that's preventing us from doing an equivalent reader in CMake here? At least that was we're not jumping between disconnected components.

Something to this tune:

```cmake
file(STRINGS versions.txt lines)
foreach(line ${lines})
# ...
```

yrashk: Also, referring to `versions.txt` without a path may backfire. Perhaps a location relative to Version.cmake (that is, the current list dir) can be better? 

What if we save `CMAKE_CURRENT_LIST_DIR` before we define the macro in a globally cached variable and then use it to`${dir}/../version.txt` to find the file?

rmodpur: > Is there anything that's preventing us from doing an equivalent reader in CMake

The bash script detects duplicate entries of an extension as well, errors out if that's the case

yrashk: I suspect this might not work as expected when extensions are built individually, as their `CMAKE_SOURCE_DIR` won't be the root.

rmodpur: The PR did pass [build-extensions-independently](https://github.com/omnigres/omnigres/actions/runs/8096610533/job/22125731018?pr=504#logs)

rmodpur: Removed the bash script and made the changes you suggested

yrashk (approved):

PR #511, opened 2024-03-05
Cut initial release of extensions

rmodpur:


versions.txt
```diff
@@ -0,0 +1,20 @@
+omni_aws=0.1.0
```
yrashk, line 1: should we include `omni` here?

rmodpur: including omni will some change CMake files as well, so doing it in another commit

yrashk (approved):

PR #517, opened 2024-03-08
Problem: Docs for omnigres extensions download is outdated

rmodpur:
Docs still reference commit SHA versioned extensions which is no longer be maintained after the move to semver releases of extensions.

Solution: Update the docs with the current method of downloading and installing omnigres extensions.


viggy28 (approved):
Looks good. Thanks for improving the on-boarding flow 🙏

PR #518, opened 2024-03-11
Problem: omnigres extension download is not straightforward

rmodpur:
Currently, downloading an omnigres extension involves few steps like downloading and copying the extension files in the right postgres directories which is not a good from user experience perspective.

Solution: Write a bash script to encapsulate all the stuff described above to simplify omnigres extension download.

### DIFF
--- a/download-omnigres-extension.sh
+++ b/download-omnigres-extension.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+
+set -e
+
+if ! which curl > /dev/null; then
+  echo "curl not found, install curl" && exit 1
+fi
+
+subcommand="$1"
+ext_name="$2"
+ext_version="$3"
+
+OS="$(uname)"
+
+if [ "$OS" = Linux ]; then
+  OS_ARCH="ubuntu-x86-64"
+elif [ "$OS" = Darwin ]; then
+    OS_ARCH="macos-arm"
+else
+  echo "extensions not available for os: $OS" && exit 1
+fi
+
+regex="PostgreSQL[[:space:]]+([0-9]+)\.[0-9]+"
+
+pg_config_version="$(pg_config --version)"
+
+if [[ "$pg_config_version" =~ $regex ]]; then
+    PG_MAJOR_VERSION="${BASH_REMATCH[1]}"
+else
+  echo "pg_config --version output doesn't match the regex: \"$regex\"" && exit 1
+fi
+
+platform_url="https://index.omnigres.com/$PG_MAJOR_VERSION/Release/$OS_ARCH"
+
+tarball_path="/tmp/$ext_name--$ext_version.tar.gz"
+untarred_dir_path="/tmp/$ext_name--$ext_version--untarred"
+
+clean_up () {
+    exit_code=$1
+
+    if [ -f "$tarball_path" ]; then
+      rm "$tarball_path"
+    fi
+
+    if [ -d "$untarred_dir_path" ]; then
+      rm -rf "$untarred_dir_path"
+    fi
+
+    exit $exit_code
+}
+
+fail () {
+  error_msg=$1
+  echo "$error_msg
+
+Usage: $0 <subcommand>
+
+Subcommands:
+  install <ext-name> <ext-version>
+    download and copy the extension files(only if it doesn't exist) to postgres directories of the given omnigres extension version
+
+
+Visit $platform_url/index.json to view available omnigres extension releases"
+
+  clean_up 1
+}
+
+case "$subcommand" in
+
+  install)
+    if [ -z "$ext_name" ] || [ -z "$ext_version" ]; then
+      fail "extension name and version should not be empty"
+    fi
+
+    if ! curl -sSL --fail-with-body "$platform_url/$ext_name/$ext_version?with-dependencies" -o "$tarball_path"; then
+      fail "extension download error: $(cat $tarball_path)"
+    fi
+
+    rm -rf "$untarred_dir_path" && mkdir "$untarred_dir_path" && tar -C "$untarred_dir_path" -xzf "$tarball_path"
+
+    control_and_sql_files_destination="$(pg_config --sharedir)/extension"
+
+    # `cp -n` (avoid copy if file exists at destination) fails in macos with non-zero exit code if file already exists
+    # avoid overwriting of files while copying because .so file of an extension may already be loaded in memory
+    # so loop through list returned by glob and copy if file doesn't exist
+
+    shopt -s nullglob # required for returning empty list in case of no matching entries for glob
+    for file in "$untarred_dir_path"/*{.sql,.control}; do
+      if [ ! -f "$control_and_sql_files_destination"/$(basename "$file") ]; then
+        cp -v "$file" "$control_and_sql_files_destination"
+      else
+        echo "skipping copy of $(basename "$file"), already exists"
+      fi
+    done
+    shopt -u nullglob # unset the above set option
+
+    if [ -d "$untarred_dir_path/lib" ]; then
+      so_files_destination="$(pg_config --pkglibdir)"
+      shopt -s nullglob
+      for file in "$untarred_dir_path"/lib/*.so; do
+        if [ ! -f "$so_files_destination"/$(basename "$file") ]; then
+          cp -v "$file"  "$so_files_destination"
+        else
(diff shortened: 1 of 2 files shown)


rmodpur:
Output on successfully running the script:
```shell
$ curl -s https://raw.githubusercontent.com/rmodpur/omnigres/ext-download-script/download-omnigres-extension.sh | bash -s omni_httpd "0.1.0"
The sql and control files copied:
'/tmp/omni_httpd--0.1.0--untarred/omni_http--0.1.0.sql' -> '/usr/share/postgresql/16/extension/omni_http--0.1.0.sql'
'/tmp/omni_httpd--0.1.0--untarred/omni_httpd--0.1.0.sql' -> '/usr/share/postgresql/16/extension/omni_httpd--0.1.0.sql'
'/tmp/omni_httpd--0.1.0--untarred/omni_types--0.1.0.sql' -> '/usr/share/postgresql/16/extension/omni_types--0.1.0.sql'
'/tmp/omni_httpd--0.1.0--untarred/omni_http--0.1.0.control' -> '/usr/share/postgresql/16/extension/omni_http--0.1.0.control'
'/tmp/omni_httpd--0.1.0--untarred/omni_http.control' -> '/usr/share/postgresql/16/extension/omni_http.control'
'/tmp/omni_httpd--0.1.0--untarred/omni_httpd--0.1.0.control' -> '/usr/share/postgresql/16/extension/omni_httpd--0.1.0.control'
'/tmp/omni_httpd--0.1.0--untarred/omni_httpd.control' -> '/usr/share/postgresql/16/extension/omni_httpd.control'
'/tmp/omni_httpd--0.1.0--untarred/omni_types--0.1.0.control' -> '/usr/share/postgresql/16/extension/omni_types--0.1.0.control'
'/tmp/omni_httpd--0.1.0--untarred/omni_types.control' -> '/usr/share/postgresql/16/extension/omni_types.control'
The so files copied:
'/tmp/omni_httpd--0.1.0--untarred/lib/omni_httpd--0.1.0.so' -> '/usr/lib/postgresql/16/lib/omni_httpd--0.1.0.so'
'/tmp/omni_httpd--0.1.0--untarred/lib/omni_types--0.1.0.so' -> '/usr/lib/postgresql/16/lib/omni_types--0.1.0.so'
The artifact format of the extension: 'omni_httpd=0.1.0#omni_types=0.1.0,omni_http=0.1.0'
```

Output if extension version doesn't exists
```shell
$ curl -s https://raw.githubusercontent.com/rmodpur/omnigres/ext-download-script/download-omnigres-extension.sh | bash -s omni_httpd "0.1.1"
curl: (22) The requested URL returned error: 404
extension download error: no extension found at the URL '/16/Release/ubuntu-x86-64/omni_httpd/0.1.1'

Usage: bash <ext-name> <ext-version>

Visit https://index.omnigres.com/16/Release/ubuntu-x86-64/index.json to view available omnigres extension releases
```

yrashk:
Few thoughts:

1. Make the script a bit more future-proof by introducing an `install` subcommand.

2. Should this script print out omni_manifest install snippet? Can also be done as an additional subcommand, in addition to doing this upon installation.

download-omnigres-extension.sh
```diff
@@ -0,0 +1,91 @@
...
+
+control_and_sql_files_destination="$(pg_config --sharedir)/extension"
+
+echo "The sql and control files copied:"
+cp --no-clobber -v "$untarred_dir_path"/*{.sql,.control} "$control_and_sql_files_destination"
+
+if [ -d "$untarred_dir_path/lib" ]; then
+  echo "The so files copied:"
+  so_files_destination="$(pg_config --pkglibdir)"
+  cp --no-clobber -v "$untarred_dir_path"/lib/*.so  "$so_files_destination"
```
yrashk, line 84: --no-clobber is not supported on macos. It's `-n` on mac

download-omnigres-extension.sh
```diff
@@ -0,0 +1,91 @@
...
+
+if [[ "$pg_config_version" =~ $regex ]]; then
+    PG_MAJOR_VERSION="${BASH_REMATCH[1]}"
+else
+  echo "pg_config --version output doesn't match the regex: \"$regex\"" && exit 1
+fi
+
+platform_url="https://index.omnigres.com/$PG_MAJOR_VERSION/Release/$OS_ARCH"
+
+tarball_path="/tmp/$ext_name--$ext_version.tar.gz"
```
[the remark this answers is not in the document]
rmodpur: should i just use the current working directory, i cleanup the files irrespective of success or failure

yrashk:
> > script a bit more future-proof by introducing an install subcommand
> 
> what do you mean? download and copy the files under an `install` subcommand ?

Yes. This opens more options down the road.

rmodpur:
@yrashk 
I see the bash script as temporary solution, planning to introduce the following function in `omni_manifest`
`omni_manifest.download(omni_manifest.requirement[]) returns omni_manifest.artifact[]` which can be used as an input to `omni_manifest.install`. What do you think ?

yrashk:
> @yrashk I see the bash script as temporary solution, planning to introduce the following function in `omni_manifest` `omni_manifest.download(omni_manifest.requirement[]) returns omni_manifest.artifact[]` which can be used as an input to `omni_manifest.install`. What do you think ?

Currently, `omni_manifest` is made in a TLE to enable its adoption across a variety of services. Adding HTTP support (through, say, httpc) there will break that. We can do fdw/dblink, chances are, it's almost universally supported? But that does increase the complexity on our side and friction on the user's side.

So it depends on how we chart its path.

download-omnigres-extension.sh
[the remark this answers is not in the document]
yrashk: Current working directory may not be available for writing. It's the usual cascade of discovery that needs to be done, I suppose? Unless we redownload it a few times and pipe to tar with `--exclude`.

download-omnigres-extension.sh
```diff
@@ -0,0 +1,123 @@
...
+    shopt -u nullglob
+
+    if [ -d "$untarred_dir_path/lib" ]; then
+      so_files_destination="$(pg_config --pkglibdir)"
+      shopt -s nullglob
+      for file in "$untarred_dir_path"/lib/*.so; do
+        if [ ! -f "$so_files_destination"/$(basename "$file") ]; then
+          cp -v "$file"  "$so_files_destination"
+        else
+          echo "skipping copy of ./lib/$(basename "$file"), already exists"
```
yrashk, line 100: `install` is also very good at this, I think.

download-omnigres-extension.sh
```diff
@@ -0,0 +1,123 @@
...
+      shopt -u nullglob
+    fi
+
+    # not printing for omni_manifest to avoid confusion
+    # as omni_manifest is reponsible for creating other extensions
+    if [ "$ext_name" != "omni_manifest" ]; then
+      artifact_format=$(grep "^${ext_name}=" $untarred_dir_path/artifacts.txt)
+
+      echo -e "\nThe omni_manifest query to create the extension:
+      select *
```
yrashk, line 112: single-line it to make it easier to copy-paste?

download-omnigres-extension.sh
yrashk: (this is not a roadblock but an improvement suggestion)

download-omnigres-extension.sh
rmodpur: I made it multi-line so that it stands out in the script output

download-omnigres-extension.sh
rmodpur: went with this approach because `cp -n` errors out in macos if file already exists at the destination,

looking at `install` man page don't see an option for copying the file only if it doesn't exist.

Avoiding overwritting of files while copying because `.so` file of an extension may already be loaded in memory thus want to avoid deleting and creating the same file

download-omnigres-extension.sh
yrashk: Standing out is a good idea, copying multi-line is just a bit harder. Not a critical thing, just a passing thought.

download-omnigres-extension.sh
yrashk: Got it! It makes sense.

rmodpur: I will add the comments

rmodpur: Added the comments

yrashk (approved):

PR #522, opened 2024-03-13
omni 0.1.1

### LINKED ISSUE #523: Problem: omni extension with different version of preloaded omni
If one version of `omni` is preloaded but another is used as an extension, there's a non-zero chance that they are incompatible, and it can thus lead to undefined behavior.

Proposed solution 1: make `omni` the preloaded library expose its version so that omni the extension can check it and error out if unsupported.

yrashk:
- [x] Ensure versions.txt gets updated with an appropriately versioned `omni` (we start with 0.1.1 but that may be not the best indication)

Closes #523


rmodpur (changes requested):
To whomsoever planning to merge this PR, please bump the version of omni in these files as well (to not break omni docker image)
https://github.com/omnigres/omnigres/pull/526/files

Going forward this will be not be required because the version of omni will be picked from `versions.txt`

yrashk:
> To whomsoever planning to merge this PR, please bump the version of omni in these files as well (to not break omni docker image) https://github.com/omnigres/omnigres/pull/526/files
> 
> Going forward this will be not be required because the version of omni will be picked from `versions.txt`

I am not sure how this will be picked up in the future based on the current source code.

rmodpur:
> > To whomsoever planning to merge this PR, please bump the version of omni in these files as well (to not break omni docker image) https://github.com/omnigres/omnigres/pull/526/files
> > Going forward this will be not be required because the version of omni will be picked from `versions.txt`
> 
> I am not sure how this will be picked up in the future based on the current source code.

By creating entrypoint files for docker image in dockerfile itself

rmodpur (approved):

PR #524, opened 2024-03-13
Problem: omni hook's wrap

yrashk:
There a few problem with this:

* It's unclear what does it do
* It does not allow us to add hooks at the end of the list, only at the beginning or both.

Solution: change it to `position` that can be bitwise-OR'd to combine "leading" and "trailing" positions.

If no position specified, assume it's "leading" to mimic the existing behavior in absence of `wrap` (which meant `false`)


extensions/omni/omni.c
```diff
@@ -482,7 +482,11 @@ static void register_hook(const omni_handle *handle, omni_hook *hook) {
   // Ensure we allocate in the top memory context as we need to keep these around forever
   MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
 
-  if (hook->wrap) {
+  if (hook->position == 0) {
```
rmodpur, line 485: can use `omni_hook_position_default` instead of `0`

rmodpur:
1. https://github.com/yrashk/omnigres/blob/4d78f4492821c74841ecb807d3396ddf342bc491/extensions/omni/omni.c#L477
shouldn't it be `hook->type < __OMNI_HOOK_TYPE_COUNT` rather than `hook->type <= __OMNI_HOOK_TYPE_COUNT`

2. How about calculating the number of entries at the beginning of the function so that we can `palloc/repalloc` only once

omni/omni/omni_v0.h
```diff
@@ -47,7 +47,7 @@ typedef struct {
 StaticAssertDecl(sizeof(omni_magic) <= UINT16_MAX, "omni_magic should fit into 16 bits");
 
 #define OMNI_INTERFACE_VERSION 0
-#define OMNI_INTERFACE_REVISION 5
+#define OMNI_INTERFACE_REVISION 6
```
rmodpur, line 50: I just realized there are two types of versions in omni:
1. `x.y.z` semver version for extension
2. `OMNI_INTERFACE_REVISION` for omni interface
Isn't it confusing because we can introducing breaking changes by bumping the second type of version without reflecting it in the extension version ?

yrashk: Yes, indeed there are two. Interface versioning implies only the shape of the interface. Works this way:

< OMNI_INTERFACE_VERSION ><OMNI_INTERFACE_REVISION> (where it is suggested to translate OMNI_INTERFACE_REVISION to a letter in the alphabet to make sure people don't treat it as a SemVer; so it'd be 1A, 2B, etc.)

The idea is that 1B can only augment 1A but can't break compatibility, where as 2A can break anything from 1*.

The only exception is VERSION 0: in 0A..Z where we can break anything.

extensions/omni/omni.c
yrashk: Good point. Will make it much more readable.

PR #525, opened 2024-03-14
Problem: Some omnigres extensions fail silently if omni is not shared preloaded

rmodpur:
Extensions like `omni_httpd` which depend on `libomni` fail silently if omni shared object library is not shared preloaded.

Solution: For extensions depending on `libomni` check if omni is shared preloaded in it's sql script to prevent creating the extension in first place if it's not preloaded.

### DIFF
--- a/cmake/PostgreSQLExtension.cmake
+++ b/cmake/PostgreSQLExtension.cmake
@@ -322,6 +322,35 @@ $<$<NOT:$<BOOL:${_ext_SUPERUSER}>>:#>superuser = ${_ext_SUPERUSER}
     endif()
     endif()
 
+    # add this if an extension depends on libomni
+    set(omni_preloaded_check [=[
+do
+\$\$
+    declare
+        present boolean;
+    begin
+        select
+            -- ensure last component of matched library is omni
+            coalesce(match[array_length(match, 1)], '') ~ 'omni--\d+\.\d+\.\d+(\.so)?$'
+        from
+            regexp_split_to_array((select
+                                       library
+                                   from
+                                       -- comma separated list of libraries
+                                       regexp_split_to_table(current_setting('shared_preload_libraries'),
+                                                             ',') as library
+                                   where
+                                       -- pick the library which ends with omni--x.y.z where .so is optional
+                                       library ~ 'omni--\d+\.\d+\.\d+(\.so)?$'), '/') as match
+        into present;
+        if not present then
+            raise exception 'extension requires omni but omni not found in shared_preload_libraries'
+                using hint = 'add omni shared object library to shared_preload_libraries';
+        end if;
+    end;
+\$\$ language plpgsql;
+]=])
+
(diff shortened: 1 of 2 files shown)


cmake/PostgreSQLExtension.cmake
```diff
@@ -353,6 +382,9 @@ for f in $(ls \"$_dir/\"*.sql | sort -V)
   do
   pushd $(pwd) >/dev/null
   cd \"${CMAKE_CURRENT_SOURCE_DIR}\"
+  $<$<IN_LIST:libomni,${_ext_DEPENDS_ON}>:cat << EOF $<ANGLE-R>$<ANGLE-R> \"$1/_$$_${NAME}--${_ext_VERSION}.sql\"
+${omni_preloaded_check}
+EOF>
```
rmodpur, line 387: checks are inserted twice in extension sql script

rmodpur:
@yrashk is there any other way to detect whether omni is shared preloaded or not ? other than checking `shared_preload_libraries` value in postgres. The issue with it is it can contain multiple libraries and there are variety of ways it can be specified like
1. `shared_preload_libraries = 'omni--0.1.0.so', 'plrust'`
2. `shared_preload_libraries = 'omni--0.1.0.so,plrust'`
which makes it tricky to check whether it contains `omni` or not

yrashk:
Another way here would be to explicitly require omni as an extension to be installed, although there is no technical need for that.

yrashk:
So, here's another take based on some older ideas.

We're looking at a type of requirement we can't specify in the extension's control file. We __can__ (with varying reliability) check for omni's presence during extension installation in the script, but that won't help us if it `omni` disappears later.

We __could__ check for [compatibly versioned] `omni` to be present by setting some flag in `_Omni_init` and checking in some function that we call in `omni_httpd` (basically, observing whether `_Omni_init` was ever called). But this doesn't help us in the case of `omni_httpd` because we have a case where if `omni` is not loaded, nothing will happen. No function will be called because the background worker has not been started.

So, here's the proposal:

Let's have a "heathcheck" view in `omni_httpd". The column can be something like: 

* name
* description
* status
* resolution
* timestamp
* additional_info

Zooming in, one particular row should be something like `omni_httpd.omni` which would (by virtue of being a view) will call into a function inside of `omni_httpd`, which will rely on a flag set by `_Omni_init` discussed above.

This way, we can observe the problem if the status of this is "failed". More health checks can and should be added.

Now, to make operations even easier, we can develop a [PL/pgSQL?] extension `omni_health` that would scan for all such views in the database (either by name pattern or by structure, or both), and `select union` from all of them, providing a centralized view into all health checks. This can be easily exposed through other means like HTTP (using omni_httpd). 

This can feed into another idea that was discussed before: a proactive alert/notification system that can monitor healthchecks (that's why `resolution` above can be important) and send "pager" notifications.

PR #527, opened 2024-03-20
Problem: Omni doesn't have support for planner hook

viggy28:
Solution: Added support for planner hook with test included.


extensions/omni/hook_harness.c
```diff
@@ -148,6 +160,14 @@ MODULE_FUNCTION void omni_check_password_hook(const char *username, const char *
                 validuntil_null);
 }
 
+MODULE_FUNCTION PlannedStmt *omni_planner_hook(Query *parse, const char *query_string,
+                                               int cursorOptions, ParamListInfo boundParams) {
+  PlannedStmt *stmt =
+      iterate_hooks(omni_hook_planner, parse, query_string, cursorOptions, boundParams)
```
yrashk, line 166: This has to be changed to `iterate_hooks(planner` (similarly to all adjacent hooks) to account for changes introduced in omni's inner workings (see 4ced20bec69b98cab68e23abc4d3b42930dffe39)

PR #535, opened 2024-03-25
Problem: eliminating bugs

yrashk:
Some bugs are not immediately observable. Often not until they exhibit themselves in production.

Solution: start using sanitizers in Debug builds

This will help us observing some of the issues earlier.

Initially, let's enable UBSan and ASan and observe how well will this work.

In order for this to happen, Postgres has to be recompiled with new flags.

Caveat for macOS: leak detector is not enabled with Apple's Clang. However, one can use Homebrew's clang (from the `llvm` package) to enable it:

```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=$(brew --prefix llvm)/bin/clang  -DCMAKE_CXX_COMPILER=$(brew --prefix llvm)/bin/clang++
```

### DIFF
--- a/cmake/FindPostgreSQL.cmake
+++ b/cmake/FindPostgreSQL.cmake
@@ -119,6 +119,8 @@ if(NOT DEFINED PG_CONFIG)
             endif()
         endif()
 
+        set(OLD_CC $ENV{CC})
+
         # Ensure the right OpenSSL gets configured
         if(DEFINED OPENSSL_ROOT_DIR)
             set(OLD_CFLAGS "$ENV{CFLAGS}")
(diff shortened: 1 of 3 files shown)


cmake/FindPostgreSQL.cmake
```diff
@@ -125,6 +125,9 @@ if(NOT DEFINED PG_CONFIG)
             set(OLD_LDFLAGS "$ENV{LDFLAGS}")
             set(ENV{CFLAGS} "${OLD_CFLAGS} -I ${OPENSSL_ROOT_DIR}/include")
             set(ENV{LDFLAGS} "${OLD_LDLAGS} -L ${OPENSSL_ROOT_DIR}/lib")
+            # This ensure whatever compiler we're using to compile Omnigres, we're going
+            # to sue the same for Postgres
```
viggy28, line 129: s/sue/use?

viggy28 (approved):

cmake/FindPostgreSQL.cmake
yrashk: yes, last working draft has this fixed already

PR #537, opened 2024-03-28
Problem: Omni doesn't have support for planner hook

viggy28:
Solution: Add support for planner hook with test included.

I merged my fork with the latest and realized the PR needs to go to `next/omni` branch. Figured a better git workflow. Hence, closing the previous PR https://github.com/omnigres/omnigres/pull/527.

### DIFF
--- a/extensions/omni/hook_harness.c
+++ b/extensions/omni/hook_harness.c
@@ -4,6 +4,7 @@
 // clang-format on
 #include <commands/user.h>
 #include <executor/executor.h>
+#include <optimizer/planner.h>
 
 #include "omni_common.h"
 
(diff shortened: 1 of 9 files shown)


extensions/omni/hook_harness.c
```diff
@@ -34,6 +35,17 @@ MODULE_FUNCTION void default_needs_fmgr(omni_hook_handle *handle, Oid fn_oid) {
           : ((needs_fmgr_hook_type)saved_hooks[omni_hook_needs_fmgr])(fn_oid);
 }
 
+MODULE_FUNCTION void default_planner(omni_hook_handle *handle, Query *parse,
+                                     const char *query_string, int cursorOptions,
+                                     ParamListInfo boundParams) {
+
+  handle->returns.PlannedStmt_value =
```
yrashk, line 42: As noted in `planner` documentation:

<blockquote>
standard_planner() scribbles on its Query input, so you'd better copy that data structure if you want to plan more than once.
</blockquote>

Should we consider giving `standard_planner` a copy of `Query` here to avoid potential problems How would we test this behavior?

extensions/omni/test/omni_test.c
```diff
@@ -136,6 +136,12 @@ void _Omni_init(const omni_handle *handle) {
                                   .fn = {.xact_callback = xact_callback}};
   handle->register_hook(handle, &xact_callback_hook);
 
+  omni_hook planner_hook = {.name = "planner_hook",
+                            .type = omni_hook_planner,
+                            .fn = {.planner = planner_hook_fn},
+                            .wrap = true};
```
rmodpur, line 142: I'm not sure whether setting `wrap` to true is required here, as I understand `wrap` will add the same hook to the beginning and end of list of a particular hook type, an example usecase of `wrap` would help to decide when to use it.

extensions/omni/hook_harness.c
viggy28: the planner is modifying `Query` significantly in `subquery_planner()`.

Shall we copy here and update the handle returns if `Query` is modified after `standard_planner()` so further hooks can use the original `Query` if they need to?

yrashk: Two thoughts here:

1. Can we document (reasonably well) what the standard planner is "scribbling" on Query?
2. Would followup hooks ever need to see scribbled version of Query per chance?

viggy28: 1. Theoretically, it's possible to track down but given the number of places & fields modified, I fear it won't cover all cases and end up not useful in practice

2. I looked at what are the hooks that gets executed after `planner`. Good news is not many. It's mainly Executor hooks and they aren't relying on `Query`.

yrashk: As for (2), I meant planner hooks that one would set up. Do we want them to see the original Query, scribbled-on Query, or be able to choose which one they want to see?

viggy28: As discussed internally, given the lack of scenarios that we could think of (and the need for it), we will keep the scribbled on query as is and it will be passed to the successive `standard_planner()` calls. 

Will add a note to make that clear.

Issue #484, opened 2024-02-07
Problem: terminating pg_yregress doesn't always kill & cleanup Postgres


yrashk:
When pg_yregress is terminated (by Ctrl-C) or due to an error, it doesn't always terminate managed instances of Postgres.

This makes running tests cleanly and without polluting processes / shared memory difficult.

Proposed solution: ensure we can always terminate Postgres that was started by pg_yregress

It has to use a robust method.

---

**Bounty size: $250**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

abhinav-upadhyay:
I would like to take a crack at it. I see that currently atexit is being used to terminate the instances but to make sure that the instances are stopped even in case of an error or forced stop using ctrl+c, we can install signal handlers. 

Is that what you had in mind?

yrashk:
@abhinav-upadhyay roughly; there may be other venues to explore here:

* what happens to those instances if pg_yregress itself crashes?
* maybe look into [POSIX process groups](https://en.wikipedia.org/wiki/Process_group)?

Also, depending on how Postgres did (it may have crashed because we're testing extensions and they may crash!), we might need to cleanup shared memory segments, IPC semaphores (ipcs/ipcrm, see https://github.com/omnigres/omnigres/wiki/macOS-Quirks,-Tips-&-Tricks#runinng-out-of-shared-memory -- this is actually not macOS specific)

abhinav-upadhyay:
Thanks for the pointers. I think for if pg_yregress itself crashes, for example due to segmentation fault - that could also be handled via the signal handler. But I will look into process groups, perhaps it is a better way to do it.

BTW, I am trying to build omnigres on my laptop. Currently seeing the following error. 

```
/usr/bin/ld: libpq/be-secure-openssl.o: in function `be_tls_open_server':
/home/abhinav/dev/omnigres/.pg/Linux-Debug/16.1/postgresql-16.1/src/backend/libpq/be-secure-openssl.c:574: undefined reference to `SSL_get1_peer_certificate'
/usr/bin/ld: libpq/be-secure-openssl.o: in function `my_SSL_set_fd':
/home/abhinav/dev/omnigres/.pg/Linux-Debug/16.1/postgresql-16.1/src/backend/libpq/be-secure-openssl.c:947: undefined reference to `ERR_new'
[...]

PR #538, opened 2024-03-30
Problem: in some circumstances managed Postgres data is not cleaned up

UkuLoskit:
Addresses https://github.com/omnigres/omnigres/issues/484

### DIFF
--- a/pg_yregress/instance.c
+++ b/pg_yregress/instance.c
@@ -204,14 +204,25 @@ connect:
   return yinstance_connect_failure;
 }
 
+void yinstance_prepare(yinstance *instance) {
+  if (instance->managed) {
+    char datadir_template[] = "pg_yregress_XXXXXX";
+    char *datadir = mkdtemp(datadir_template);
+    if (datadir == NULL) {
+      fprintf(stderr, "Failed to create temporary directory: %s\n", strerror(errno));
+      return;
+    }
+    char *heap_datadir = strdup(datadir);
+    instance->info.managed.datadir = (iovec_t){.base = heap_datadir, .len = strlen(heap_datadir)};
+  }
+}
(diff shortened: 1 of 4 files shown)


pg_yregress/instance.c
```diff
@@ -204,14 +204,25 @@ yinstance_connect_result yinstance_connect(yinstance *instance) {
...
+      fprintf(stderr, "Failed to create temporary directory: %s\n", strerror(errno));
+      return;
+    }
+    char *heap_datadir = strdup(datadir);
+    instance->info.managed.datadir = (iovec_t){.base = heap_datadir, .len = strlen(heap_datadir)};
+  }
+}
 
-    char datadir[] = "pg_yregress_XXXXXX";
-    mkdtemp(datadir);
```
UkuLoskit, line 211: previously `mkdtemp`'s return value was not used which seems like a bug. `mkdtemp` can fail, in which case the return value would be `NULL`

pg_yregress/instance.c
```diff
@@ -204,14 +204,25 @@ yinstance_connect_result yinstance_connect(yinstance *instance) {
   return yinstance_connect_failure;
 }
 
-void yinstance_start(yinstance *instance) {
+void yinstance_prepare(yinstance *instance) {
   if (instance->managed) {
+    char datadir_template[] = "pg_yregress_XXXXXX";
+    char *datadir = mkdtemp(datadir_template);
+    if (datadir == NULL) {
+      fprintf(stderr, "Failed to create temporary directory: %s\n", strerror(errno));
```
UkuLoskit, line 212: TODO: should be a boolean function to indicate whether this succeeded or we could bail out of the whole process as well

Issue #211, opened 2023-06-18
Problem: libh2o aborts process on fatal errors


yrashk:
This would mean that during these errors, entire Postgres process will be ungracefully terminated

Proposed solution: replace fatal error termination as pre https://github.com/h2o/h2o/issues/3244 or find a similar solution

yrashk:
Progress by @rmodpur https://github.com/h2o/h2o/pull/3323

@rmodpur, do you want me to assign this bounty to you?

I guess now we need to be able to supplement our own h2o_fatal handling.

rmodpur:
> do you want me to assign this bounty to you?

Yes

rmodpur:
Assuming we can override the default `h2o_fatal` what would it look like ? It's replacement will also have to ultimately exit right? Because we can't make any assumption about where `h2o_fatal` will be used by the `h2o` library so things like logging and returning is not an option

rmodpur:
Thinking out loud for options other than just exiting
1. If some `h2o` function is running in the context of user defined postgres function, `h2o_fatal` leads to an exception (unsure about languages other than `plpgsql`)
2. If an `h2o` function running in context of background worker calls `h2o_fatal` it will lead to restart of the background worker (loss of that particular background worker state)

yrashk:
I think you are in the right ballpark. 

1. I think it can just be a regular `ereport` and thus it will be work in any language, but we need to make the right determination here whether it's an `ERROR` or `FATAL`
2. Yes, or do the event loop thread restart, but that's more difficult to get right with regard to message queues and locks so it may indeed be easier to log the error and restart the whole background worker. **NB**: we can't directly access any Postgres APIs from the event loop thread.
2.1. While the above (restarting the whole background worker) seems to be a good option for the current architecture, the upcoming generalized workers architecture will likely have a singular background worker for all non-Postgres workloads (http, timer, etc.) and restarting the whole background worker would be a bit too impactful. But perhaps at that point, we can design event loop threads to be easily restartable individually to minimize the impact. That said, we need to research and see if that's going to be ok with how libh2o is implemented.

mthom:
I'll take this bounty if it's free.

yrashk:
@mthom assigned you! Thank you.

rishi-jat:
@yrashk can I work on this issue ?

Kiki-bo-zhang:
Hi,

[...]

PR #539, opened 2024-03-31
Problem: h2o_fatal calls abort() by default and cannot be overridden

mthom:
Solution: h2o newly supports overriding the h2o_fatal error handler by expanding to a function pointer call. Update the h2o version so that omnigres can take advantage of this recent change.

Also update the `h2o_quic_init_context` call in omni_httpc, whose prototype was recently changed to include a new argument, whose value in omnigres is now NULL (#211)

I'm not sure if this PR is a solution to #211, since it does not itself overload the `h2o_fatal` handler, it only makes it possible. This is because the comments of #211 are ambivalent on what the overload should do.

### DIFF
--- a/cmake/FindH2O.cmake
+++ b/cmake/FindH2O.cmake
@@ -6,6 +6,6 @@ list(APPEND _h2o_options "WITH_MRUBY OFF")
 list(APPEND _h2o_options "DISABLE_LIBUV ON")
 
 cmake_policy(SET CMP0042 NEW)
-CPMAddPackage(NAME h2o GIT_REPOSITORY https://github.com/h2o/h2o GIT_TAG 579135a VERSION 2.3.0-579135a OPTIONS "${_h2o_options}")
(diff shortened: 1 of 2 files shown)


yrashk:
It appears that there's a problem with the change in h2o on macOS/clang:

```
/Users/runner/_work/omnigres/omnigres/cpm_modules/h2o/4a28805e2ec67783767fc04cf4201af69a683c42/include/h2o/memory.h:187:8: error: '_Noreturn' can only appear on functions
extern H2O_NORETURN void (*h2o__fatal)(const char *file, int line, const char *msg, ...) __attribute__((format(printf, 3, 4)));
       ^
/Users/runner/_work/omnigres/omnigres/cpm_modules/h2o/4a28805e2ec67783767fc04cf4201af69a683c42/include/h2o/memory.h:58:22: note: expanded from macro 'H2O_NORETURN'
#define H2O_NORETURN _Noreturn
```

mthom:
Not sure what we can do about that.

PR #545, opened 2024-04-06
Problem: omni_httpd worker sporadically stops listening

yrashk:
This is manifesting in scenarios like recursive HTTP calls (HTTP handler calling back into the same instance of omni_httpd) and having omni_httpd not receive an opportunity to accept an incoming connection.

Solution: fix a bug in socket read resumption
    
Previously, what we attempted in ae72953f406d8716741f0c24f08b0e51321ca20a was subtly (and critically)
wrong. Specifically, we were trying to resume reading from a socket that was holding the worker busy. 

However, that's not the socket that needs resumption, as we actually need to resume all other sockets that
were, in fact, stopped.


rmodpur:
I tried upgrading from omni_httpd 0.1.1 to omni_httpd 0.1.2, listeners present in `omni_httpd.listeners` table were not created, even restarting doesn't help

rmodpur (approved):

PR #549, opened 2024-04-09
Problem: omni_python doesn't have a dbapi library

JoshInnis:
When using omni_python the option to use Python's DB API interface is not easily available. Users either have to convert their application to use plpy or use psycopg over the wire.

Solution: Using petere/plpydbapi as a template implement a library in omni_python that supports this interface. The library is currently called dbapi.


languages/python/omni_python/src/omni_python/dbapi.py
```diff
@@ -0,0 +1,361 @@
+"""
```
yrashk, line 1: I think it would be prudent to include something like: "Based on ...." (as we're expected to evolve this) and include license terms

https://github.com/petere/plpydbapi/blob/master/LICENSE

yrashk (commented):
I think this is okay (with minor comment supplied) for a first go as it does enable basic DBAPI. 

What do you think of using something like this https://github.com/baztian/dbapi-compliance to test compliance?

languages/python/omni_python/src/omni_python/dbapi.py
JoshInnis: Added the original license

PR #558, opened 2024-04-11
Problem: CLion sometimes doesn't format properly

viggy28:


HACKING.md
```diff
@@ -32,3 +32,5 @@ You can also set up CLion to do this automatically for you (it is done this corr
 ![CLion setting for ClangFormat](docs/clion_clang_format.png)
 
 ![CLion setting for auto-save formatting](docs/clion_format_on_save.png)
+
+If CLion doesn't format automatically, you can use [clion-format](https://clang.llvm.org/docs/ClangFormat.html) CLI to format it.
```
yrashk, line 36: it's clang-format and we have a build target for it:

```
make fix-format
```

which is already described a few lines above https://github.com/omnigres/omnigres/blob/master/HACKING.md#source-code-formatting

viggy28: My bad. That's suffice.

Issue #543, opened 2024-04-02
Problem: overriding `postgres` symbols


yrashk:
As discovered during a debugging session on #537, there's a chance for a subtle problem when a Postgres symbol is overridden by an extension. Specifically, if a symbol is defined in the extension and `postgres`, the linker may pick `postgres`'s version of the symbol over the one in the extension

On macOS/clang it seems to compile and link correctly (as we'd expect it) but on Linux/gcc it didn't.

Proposed solution: run an intersection of `nm -g --defined-only` (or `nm -g -U` on macOS, it seems)
on both the compiled extension and `postgres`.

If there's any intersection, error out.

---

**Bounty size: $100**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

UkuLoskit:
Hey, @yrashk ! Could you assign it to me? Seems straight-forwad enough.

yrashk:
Here you go!

PR #566, opened 2024-04-16
Problem: overriding postgres symbols

UkuLoskit:
Addresses #543 

Solution: add a target that checks if Postgres and a Postgres extension have conflicting symbols. If conflicting symbols are detected the build shall fail.


cmake/PostgreSQLExtension.cmake
```diff
@@ -255,6 +255,16 @@ function(add_postgresql_extension NAME)
                 PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
     endif()
 
+    # Check that the extension has no conflicting symbols with the Postgres binary
+    # otherwise the linker might use the symbols from Postgres
+    find_package(Python COMPONENTS Interpreter REQUIRED)
+    execute_process(COMMAND dirname ${PG_CTL} OUTPUT_VARIABLE POSTGRES_BINARY)
+    get_filename_component(POSTGRES_BINARY ${PG_CTL}/../postgres ABSOLUTE)
+    set(SO_FILE "${CMAKE_CURRENT_BINARY_DIR}/${_ext_TARGET}${_suffix}")
+    add_custom_target(omni_check_symbol_conflict_${_ext_TARGET} ALL
```
UkuLoskit, line 264: ALL is probably not the target to create the dependency on... need a bit help with this.

cmake/PostgreSQLExtension.cmake
```diff
@@ -255,6 +255,16 @@ function(add_postgresql_extension NAME)
                 PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
     endif()
 
+    # Check that the extension has no conflicting symbols with the Postgres binary
+    # otherwise the linker might use the symbols from Postgres
+    find_package(Python COMPONENTS Interpreter REQUIRED)
+    execute_process(COMMAND dirname ${PG_CTL} OUTPUT_VARIABLE POSTGRES_BINARY)
+    get_filename_component(POSTGRES_BINARY ${PG_CTL}/../postgres ABSOLUTE)
```
UkuLoskit, line 262: couldn't find a nicer way of obtaining the postgres binary path

yrashk: FindPostgreSQL.cmake defines `PG_BINARY`

cmake/PostgreSQLExtension.cmake
yrashk: Target file can be obtained more easily (I think) with `$<TARGET_FILE:target>`

As for the dependency, maybe `(package_${_ext_TARGET}_extension`? It's not going to get during the build process, though but packaging. 🤔

cmake/check_symbol_conflict.py
```diff
@@ -0,0 +1,35 @@
+import platform
```
yrashk, line 1: This is totally nitpicking (not critical): I think it can be solved relatively simply with awk/sort/comm. Again, it is really not critical at all, rather just a thought. Chances are, Python version is more readable than that.

yrashk:
👋 Uku, I was wondering if you saw the comments I left above. Let me know if this helps!

UkuLoskit:
> 👋 Uku, I was wondering if you saw the comments I left above. Let me know if this helps!
Yes, I did. Since I was not sure of the solution, I got a bitstuck on it

yrashk:
I did a manual review by trying to cause a problem in one of the extensions (by adding `void standard_planner() {}`), but it didn't work.

There seems to be a problem with the dependency ordering of CMake targets. I've addressed this (and minor visual elements in the Python script) in this patch:

```diff
diff --git a/cmake/PostgreSQLExtension.cmake b/cmake/PostgreSQLExtension.cmake
index b38d789..1fd1508 100644
--- a/cmake/PostgreSQLExtension.cmake
+++ b/cmake/PostgreSQLExtension.cmake
@@ -375,6 +375,7 @@ $command $@
     if(_ext_SOURCES)
         add_custom_target(package_${_ext_TARGET}_extension
                 WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                DEPENDS omni_check_symbol_conflict_${_ext_TARGET}
                 COMMAND
                 ${CMAKE_COMMAND} -E copy_if_different
                 "${CMAKE_CURRENT_BINARY_DIR}/$<TARGET_FILE_NAME:${_ext_TARGET}>"
@@ -382,9 +383,9 @@ $command $@
         # Check that the extension has no conflicting symbols with the Postgres binary
         # otherwise the linker might use the symbols from Postgres
         find_package(Python COMPONENTS Interpreter REQUIRED)
-        add_custom_target(omni_check_symbol_conflict_${_ext_TARGET} package_${_ext_TARGET}_extension
+        add_custom_target(omni_check_symbol_conflict_${_ext_TARGET}
                 WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-                COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/check_symbol_conflict.py ${PG_BINARY} "${CMAKE_CURRENT_BINARY_DIR}/$<TARGET_FILE:${_ext_TARGET}>")
+                COMMAND ${Python_EXECUTABLE} ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/check_symbol_conflict.py ${PG_BINARY} "$<TARGET_FILE:${_ext_TARGET}>")
     endif()
 
     add_custom_target(package_${_ext_TARGET}_migrations
diff --git a/cmake/check_symbol_conflict.py b/cmake/check_symbol_conflict.py
index 523889a..c6ba382 100644
--- a/cmake/check_symbol_conflict.py
+++ b/cmake/check_symbol_conflict.py
@@ -30,6 +30,7 @@ if __name__ == "__main__":
     symbols2 = get_symbols(binary2)
     intersection = symbols1 & symbols2
     if symbols1 & symbols2:
-        sys.stderr.write("Found symbols that are present in both binaries:\n")
-        sys.stderr.write("\n".join(intersection))
+        sys.stderr.write("Conflicting symbols with `postgres`:\n  ")
+        sys.stderr.write("\n  ".join(intersection))
+        sys.stderr.write("\n")
         sys.exit(1)
```

It did work as expected after this patch. Feel free to re-use bits of this diff above!

P.S. Too bad we don't (yet) have tests for the build harnessing itself :]

UkuLoskit:
@yrashk thanks! I did not really know how to improve the script since you provided the diff :)

yrashk:
Looks like we've got another symbol detected in CI!

```
Conflicting symbols with `postgres`:
[100%] Built target package_omni_sql_extension
  _xpath
```

yrashk:
Now it seems to fail on `omni` extension:

```
Conflicting symbols with `postgres`:
  dshash_dump
  dshash_detach
  dshash_create
  dshash_attach
  dshash_memcmp
  dshash_find
  dshash_destroy
  dshash_get_hash_table_handle
  dshash_memhash
  dshash_delete_entry
  dshash_find_or_insert
  dshash_release_lock
  dshash_delete_key
 ```

Which is interesting as I can't yet reproduce this locally. The check in omni to include some of dshash functionality is this:


```c
#if PG_MAJORVERSION_NUM < 15
```

That said, it is indeed not exactly right that we include all functions in versions below 15 as these symbols are present there anyway. But I don't yet understand why it fails on 15+, too.

UkuLoskit:
glad to see that it actually works and is finding some (potential?) issues

yrashk:
Can't reproduce locally on Linux either, yet:

```
$ nm -g --defined-only ./extensions/omni/omni--0.1.2.so | grep dshash | wc -l
0
```

🤔

yrashk:
Symbol overlap only happened on Postgres 14 and 13—fixed that.

yrashk (approved):

PR #567, opened 2024-04-19
omni_schema is still not solving all problems (🚀 omni_schema 0.2.0)

yrashk:


extensions/omni_schema/tests/assemble_schema.yml
```diff
@@ -0,0 +1,283 @@
+$schema: "https://raw.githubusercontent.com/omnigres/omnigres/master/pg_yregress/schema.json"
+instance:
+  init:
+    - create extension omni_schema cascade
+    - create extension dblink cascade
```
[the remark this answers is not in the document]
yrashk: Yeah, I already did! Thank you.

extensions/omni_schema/src/assemble_schema.sql
```diff
@@ -0,0 +1,271 @@
...
+    for rec in select distinct extension
+               from omni_schema.languages
+               union
+               select distinct file_processor_extension as extension
+               from omni_schema.languages
+               union
+               select distinct processor_extension as extension
+               from omni_schema.auxiliary_tools
+        loop
+            if exists (select from pg_extension where extname = rec.extension) then
```
rmodpur, line 55: 1. `pg_extension` lists installed extensions so shouldn't it be `not exists` ?
2.  couple of issues with creating all language extensions:
  a. not all language extensions will be used by schema files
  b. not all language extensions may be available for installation

yrashk: 1. This translates to "if this extension installed locally, then we provision it remotely"
2a. True. Temporary overhead to achieve current goals.
2b. True for truly remote dbs, but arguably not true for the local cluster, where the current use case is.

PR #586, opened 2024-07-01
Problem: using `pg_proc` to detect extensions

yrashk:
We detect extensions through references of C-language functions in `pg_proc`.

This generally works for extensions that expose some functions, but doesn't work for those that just do some work on initialization (background worker startup, etc.)

Solution: switch to `pg_extension` to detect extensions directly

One potential downside of this approach is that we won't support extensions with more than one shared object.

### DIFF
--- a/extensions/omni/init.c
+++ b/extensions/omni/init.c
@@ -28,10 +28,6 @@ static shmem_startup_hook_type saved_shmem_startup_hook;
 static void shmem_request();
 static void shmem_hook();
 
-void procoid_syscache_callback(Datum arg, int cacheid, uint32 hashvalue) {
-  backend_force_reload = true;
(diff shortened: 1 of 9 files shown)


extensions/omni/init.c
```diff
@@ -149,7 +149,10 @@ void _PG_init() {
   // This function may result in a FATAL error if we hit the limit of
   // syscache callbacks (`MAX_SYSCACHE_CALLBACKS`, 64). However, since we're in
   // postmaster: 1) it is okay to crash here 2) it is much less likely to happen at this time.
+  // ---
+  // NOTE: this is outdated
```
yrashk, line 153: This is to be resolved.

PR #587, opened 2024-07-02
Problem: omni_httpd looping infinitely

steve-chavez:
omni_httpd master and its workers loop infinitely when postgres `max_worker_processes` is 8 (the default).

This happens because:

- `omni_httpd.http_workers` defaults to the number of cores (16 in this case).
- The `RegisterDynamicBackgroundWorker` call fails because there are no background worker slots available (limited by `max_worker_processes`).
- The httpd master gets stuck busy-waiting for all workers to start their loops [1].
- The workers get stuck retrying the EAGAIN result of their socket [2].

Solution: handle the failed `RegisterDynamicBackgroundWorker` result.

This function returns false when the background worker can't start.

[1]: https://github.com/omnigres/omnigres/blob/9c2bcc6c9aa804a6b7e209a02d218449342aa1a7/extensions/omni_httpd/master_worker.c#L523
[2]: https://github.com/omnigres/omnigres/blob/9c2bcc6c9aa804a6b7e209a02d218449342aa1a7/extensions/omni_httpd/http_worker.c#L626

### DIFF
--- a/extensions/omni_httpd/master_worker.c
+++ b/extensions/omni_httpd/master_worker.c
@@ -511,15 +511,34 @@ void master_worker(Datum db_oid) {
       strncpy(worker.bgw_library_name, MyBgworkerEntry->bgw_library_name, BGW_MAXLEN - 1);
       for (int i = 0; i < *num_http_workers; i++) {
         BackgroundWorkerHandle *handle;
-        RegisterDynamicBackgroundWorker(&worker, &handle);
-        pid_t worker_pid;
-        if (WaitForBackgroundWorkerStartup(handle, &worker_pid) == BGWH_POSTMASTER_DIED) {
-          goto terminate;
+        if (RegisterDynamicBackgroundWorker(&worker, &handle)) {
+          pid_t worker_pid;
+          BgwHandleStatus status = WaitForBackgroundWorkerStartup(handle, &worker_pid);
+
+          switch (status) {
+          case BGWH_POSTMASTER_DIED:
+            goto terminate;
+          case BGWH_STARTED: {
+            cvec_bgwhandle_push(&http_workers, handle);
+            break;
+          }
+          default:
+            ereport(ERROR, (errcode(ERRCODE_INSUFFICIENT_RESOURCES),
+                            errmsg("Could not start %s", worker.bgw_name)));
+          }
+        } else {
+          ereport(WARNING, errcode(ERRCODE_INSUFFICIENT_RESOURCES),
+                  errmsg("Could not start %s", worker.bgw_name),
+                  errdetail("omni_httpd.http_workers = %d is higher than max_worker_processes = %d",
+                            *num_http_workers, max_worker_processes),
+                  errhint("Consider increasing max_worker_processes"));
         }
-        cvec_bgwhandle_push(&http_workers, handle);
       }
+
       // Ensure all workers have indeed started their loops
       uint32 expected = cvec_bgwhandle_size(&http_workers);
+      ereport(LOG, errmsg("Started %d omni_httpd workers", expected));
+
       while (!pg_atomic_compare_exchange_u32(semaphore, &expected, 0)) {
         expected = cvec_bgwhandle_size(&http_workers);
         HandleMainLoopInterrupts();
--- a/extensions/omni_httpd/omni_httpd.c
+++ b/extensions/omni_httpd/omni_httpd.c
@@ -166,6 +166,11 @@ void _Omni_init(const omni_handle *handle) {
 #ifdef _SC_NPROCESSORS_ONLN
   default_num_http_workers = sysconf(_SC_NPROCESSORS_ONLN);
 #endif
+  // considering the omni_httpd master worker and the "omni startup" worker used by omni TODO:
+  // revisit once omni_workers is set.
+  int threshold = max_worker_processes - 2;
+  if (default_num_http_workers > threshold)
+    default_num_http_workers = threshold;
 
   omni_guc_variable guc_num_http_workers = {
       .name = "omni_httpd.http_workers",
--- a/extensions/omni_httpd/tests/limited_max_workers.yml
+++ b/extensions/omni_httpd/tests/limited_max_workers.yml
@@ -1,0 +1,23 @@
+$schema: "https://raw.githubusercontent.com/omnigres/omnigres/master/pg_yregress/schema.json"
+instance:
+  config:
+    shared_preload_libraries: */env/OMNI_SO
+    max_worker_processes: 3
+  init:
+  - create extension omni_httpd cascade
+  - create extension omni_httpc cascade
+  - call omni_httpd.wait_for_configuration_reloads(2);
+
+tests:
+- name: should respond successfully
+  query: |
+    select
+      status,
+      error
+    from
+      omni_httpc.http_execute(
+        omni_httpc.http_request('http://127.0.0.1:' || (select effective_port from omni_httpd.listeners) || '/')
+      );
+  results:
+    - status: 200
+      error: null
--- a/extensions/omni_httpd/CHANGELOG.md
+++ b/extensions/omni_httpd/CHANGELOG.md
@@ -12,6 +12,9 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 * omni_httpd may crash under certain circumstances during early
   startup [#551](https://github.com/omnigres/omnigres/pull/551), [#556](https://github.com/omnigres/omnigres/pull/556)
 
+* omni_httpd master and its workers loop infinitely when postgres `max_worker_processes` is less than `omni_httpd.http_workers`.
+  [#587](https://github.com/omnigres/omnigres/pull/587).
+
 ## [0.1.2] - 2024-04-07
 
 ### Fixed
--- a/extensions/omni_httpd/docs/intro.md
+++ b/extensions/omni_httpd/docs/intro.md
@@ -154,6 +154,6 @@ from
 
 `omni_httpd` can be configured with the following PostgreSQL configuration variables:
 
-* `omni_httpd.http_workers` to configure the number of http workers, defaults to number of cpus
+* `omni_httpd.http_workers` to configure the number of http workers. It defaults to the number of cpus and adjusts if this is higher than what [max_worker_processes](https://www.postgresql.org/docs/current/runtime-config-resource.html#GUC-MAX-WORKER-PROCESSES) allows.
 * `omni_httpd.temp_dir` to set the temporary directory for `omni_httpd` files like unix domain sockets, defaults
   to `/tmp`


extensions/omni_httpd/tests/limited_max_workers.yml
```diff
@@ -0,0 +1,18 @@
...
+
+tests:
+- name: should respond successfully
+  query: |
+    select
+      status
+    from
+      omni_httpc.http_execute(omni_httpc.http_request('http://localhost:8080'));
+  results:
+    - status: 200
```
steve-chavez, line 18: This returns `null` randomly for me locally. Maybe I'm missing some type of `wait` signal.

extensions/omni_httpd/master_worker.c
```diff
@@ -511,12 +511,15 @@ void master_worker(Datum db_oid) {
...
-        if (WaitForBackgroundWorkerStartup(handle, &worker_pid) == BGWH_POSTMASTER_DIED) {
-          goto terminate;
+        if (RegisterDynamicBackgroundWorker(&worker, &handle)) {
+          pid_t worker_pid;
+          if (WaitForBackgroundWorkerStartup(handle, &worker_pid) == BGWH_POSTMASTER_DIED) {
+            goto terminate;
+          }
+          cvec_bgwhandle_push(&http_workers, handle);
+        } else {
+          ereport(WARNING, errmsg("Could not start %s", worker.bgw_name));
```
steve-chavez, line 521: Could the `RegisterDynamicBackgroundWorker` fail for other cases except the `max_worker_processes` one?

If not, it would be good for operators to put a HINT here, like:
```
HINT: Ensure omni_httpd.http_workers doesn't surpass max_worker_processes
```

extensions/omni_httpd/tests/limited_max_workers.yml
yrashk: It is an annoying known issue.

Currently solved in other tests with something like:

```yaml
- call omni_httpd.wait_for_configuration_reloads(2)
```

in the `init` section.

Basically, it's a timing issue – we are trying to test things before omni_httpd is able to ingest the new configuration, which is currently done post-transaction using LISTEN/NOTIFY. The waiting hack is ensuring the configuration has been reloaded. (2) is because of, well, how many times this happens in our current implementation. 

I have not spent much time fixing this anticipating work on the updated workers architecture.

extensions/omni_httpd/master_worker.c
yrashk: It "**typically** indicates that no background worker slots are currently available" (from the documentation)

Other cases that can be immediately seen:

* Executed in postmaster. Can't happen here.
* "If this is a parallel worker, check whether there are already too many parallel workers; if so, don't register another one.". This is not a parallel worker


So far, it seems that it is fairly plausible that this would be the only case and HINT would be appropriate.

yrashk (approved):
Minor: a CHANGELOG "Fixed" entry would be useful

Additional thought: 

Would it make any sense to protect the `while` loop below (and perhaps similar ones around) to re-consult with background worker status if we can't get them to all come up to make sure we're expecting the right number to come up.

That is to say that it probably would make sense to ensure there are reasonable exit paths in the loops that prone to become infinite, not just before.

extensions/omni_httpd/tests/limited_max_workers.yml
steve-chavez: Weirdly enough it was the `localhost`, with that I got `error: connection refused` randomly. 

That happens even with `omni_httpd.wait_for_configuration_reloads(2)`.

For now I've just changed `localhost` to `127.0.0.1`.

yrashk: I might be aware of this `localhost` issue, too! It has to do with h2o's resolver, which would randomly pick whatever host would resolve to. If your localhost also resolves to an IPv6 address, for example, and we're not listening there, it won't connect.

extensions/omni_httpd/master_worker.c
steve-chavez: Done. The message now looks like:

```
WARNING:  Could not start omni_httpd worker
DETAIL:  omni_httpd.http_workers = 16 is higher than max_worker_processes = 3
HINT:  Consider increasing max_worker_processes
```

It might seem like bikeshedding but from what I've seen good error messages help a lot. I think they fit in the D3X concept too.

extensions/omni_httpd/master_worker.c
```diff
@@ -511,12 +511,19 @@ void master_worker(Datum db_oid) {
...
+          pid_t worker_pid;
+          if (WaitForBackgroundWorkerStartup(handle, &worker_pid) == BGWH_POSTMASTER_DIED) {
+            goto terminate;
+          }
+          cvec_bgwhandle_push(&http_workers, handle);
+        } else {
+          ereport(WARNING,
+              errcode(ERRCODE_INSUFFICIENT_RESOURCES),
+              errmsg("Could not start %s", worker.bgw_name),
+              errdetail("omni_httpd.http_workers = %d is higher than max_worker_processes = %d", *num_http_workers, max_worker_processes),
```
yrashk, line 524: Great! 👍 

Nitpick: since we know the numbers, why don't we check if http_workers is higher than allowed before issuing a WARNING.

Thinking a bit further, we can also look at addressing it early in `master_worker` – should we even try to allow num_http_workers to be higher than max_worker_process - 1 (at least one more is necessary  for the master worker itself). Note that this will not fully prevent the issue as the total number of workers in the cluster is out of our control and is subject to change. But it may prevent the problem cascading here for obvious clear-cut cases like 8/16.

extensions/omni_httpd/master_worker.c
yrashk: Don't think this is bikeshedding. Good errors are important. I also left some minor commentary there, we can look at it now or as a follow-up improvement PR.

extensions/omni_httpd/master_worker.c
steve-chavez: Yeah, I think that would be better.

Btw, shouldn't the defaults be conservative? For example Nginx defaults to 1 worker process https://nginx.org/en/docs/ngx_core_module.html#worker_processes and can be tuned to `auto` for consuming all available CPUs.

Wouldn't it make sense for omni_httpd to be `1` by default and use `0` as it's auto mode?

steve-chavez:
> Would it make any sense to protect the while loop below (and perhaps similar ones around) to re-consult with background worker status if we can't get them to all come up to make sure we're expecting the right number to come up.

Looks like the status won't help. Once `RegisterDynamicBackgroundWorker` fails the `WaitForBackgroundWorkerStartup` function returns 0, which means `BGWH_STARTED`:

```c
typedef enum BgwHandleStatus
{
	BGWH_STARTED,				/* worker is running */
	BGWH_NOT_YET_STARTED,		/* worker hasn't been started yet */
	BGWH_STOPPED,				/* worker has exited */
	BGWH_POSTMASTER_DIED		/* postmaster died; worker status unclear */
} BgwHandleStatus;
```

(Maybe that enum should have started at 1)

> That is to say that it probably would make sense to ensure there are reasonable exit paths in the loops that prone to become infinite, not just before.

For avoiding the looping in `accept_fds`/http_worker we could check that the master worker started its socket loop, but I think just handling the failed dynamic bg worker start on the master is enough/simpler.

extensions/omni_httpd/master_worker.c
steve-chavez: Strange, I'm finding that `RegisterDynamicBackgroundWorker` always manages to start `max_worker_processes - 2`, not `max_worker_processes - 1`. 

It must be considering another worker besides the `omni_httpd` one?

steve-chavez: Ah, there's an additional `omni_startup` worker:
```
2024-07-06 12:51:10.665 -05 [607820] FATAL:  Can't register a dynamic background worker
2024-07-06 12:51:10.667 -05 [607814] LOG:  background worker "omni startup" (PID 607820) exited with exit code 
```

steve-chavez: For now I've harcoded the `2` and left it as TODO, assuming `omni_workers` might be able to handle that better.

This way at least we don't log WARNINGs unnecessarily.

extensions/omni_httpd/master_worker.c
```diff
@@ -511,15 +511,34 @@ void master_worker(Datum db_oid) {
...
+          switch (status) {
+          case BGWH_POSTMASTER_DIED:
+            goto terminate;
+          case BGWH_STARTED: {
+            cvec_bgwhandle_push(&http_workers, handle);
+            break;
+          }
+          default:
+            ereport(ERROR, (errcode(ERRCODE_INSUFFICIENT_RESOURCES),
+                            errmsg("Could not start %s", worker.bgw_name)));
```
steve-chavez, line 527: I've also added this case considering that auto_prewarm does the same:
https://github.com/postgres/postgres/blob/dc0178a7fae05915ca3ae3f60ab89dad15dd6d2e/contrib/pg_prewarm/autoprewarm.c#L830-L841

steve-chavez:
One thing that could be further improved is that this log can appear:
```
2024-07-06 13:37:34.195 -05 [656567] LOG:  omni_httpd: Established listener on 8080
```

And there are no http workers to serve the requests. But I think that can be fixed in another PR.

extensions/omni_httpd/omni_httpd.c
```diff
@@ -166,6 +166,11 @@ void _Omni_init(const omni_handle *handle) {
 #ifdef _SC_NPROCESSORS_ONLN
   default_num_http_workers = sysconf(_SC_NPROCESSORS_ONLN);
 #endif
+  // considering the omni_httpd master worker and the "omni startup" worker used by omni TODO:
+  // revisit once omni_workers is set.
+  int threshold = max_worker_processes - 2;
```
yrashk, line 171: I think we can't reasonably predict how many workers we can expect as this is subject to all installed extensions and their whims. What do you think?

yrashk (approved):

extensions/omni_httpd/omni_httpd.c
steve-chavez: Yes, correct. I believe the simplest way to solve this for now is to change the default http workers to 1. And then provide the `0` value as `auto`. Then if we emit warnings is reasonable to say that auto tried to give the best perf but it couldn't because there are no available workers.

That way at least we don't emit warnings in the default case.

What do you think?

yrashk: My primary concern than low number of workers result in poor performance, and 1 results in sequential processing of requests, which is... sad.

I wish there was a way to tell postgres that some workers can be easily scaled down under demand (other extensions wanting to register workers)

I'd suggest keeping what we have here and iterating further separately.

yrashk: My primary concern than low number of workers result in poor performance, and 1 results in sequential processing of requests, which is... sad.

I wish there was a way to tell postgres that some workers can be easily scaled down under demand (other extensions wanting to register workers)

I'd suggest keeping what we have here and iterating further separately.

Issue #407, opened 2023-12-20
Problem: omni_var's scope is limited


yrashk:
It only provides transactional variables. This is useful, but only to a point.

Proposed solution: add backend (session) and durably scoped variables

It's fairly trivial to make variables session-scoped. In that case, we need not enforce transactional boundaries like we do now. Durable variables are more interesting. We could store them as their binary encoding (if available; textual encoding otherwise) in a table alongside the type name.

PR #589, opened 2024-07-05
Problem: sometimes transactional variables are insufficient (🚀 omni_var 0.2.0)

yrashk:
They work well when we handle requests from within Postgres. However, imagine if one is trying to use RLS from a backend and wants to identify the user? Holding a transaction may prove less than efficient.

Solution: introduce session-level variables

They behave very similarly to transactional variables but ignore transactional and sub-transactional boundaries.

This is still not quite perfect as the backend needs to ensure that connections are not released back to the pool before the operation is completed and the variables needs to be set to an "invalid" value upon the return. This will not well at all in statement pooling.


yrashk:
An alternative API I was thinking of was:

```postgresql
set('var', val, scope => 'session')
```

with a default of `transaction` scope.

I think it's going to be just a tiny bit slower (the implementation would have to check extra argument), and I am not sure if having the same entry point is in any way beneficial from the API standpoint. 🤔

yrashk:
Closes #407

Issue #585, opened 2024-07-01
Sporadic CI failure shutting down omni_httpd


yrashk:
Do you enjoy debugging concurrency issues? This is likely one of those!

See #584, up to 87ea87f8a1a906306668ca1a3e29098c91346bf6

Sporadically, some of the "Extension scripts" jobs fail with this:

```
2024-06-30 23:16:20.247 UTC [29589] LOG:  omni_httpd: Established listener on 8080
2024-06-30 23:16:20.253 UTC [29575] LOG:  received smart shutdown request
waiting for server to shut down....2024-06-30 23:16:20.255 UTC [29589] LOG:  Terminated omni_httpd
2024-06-30 23:16:20.256 UTC [29575] LOG:  background worker "logical replication launcher" (PID 29583) exited with exit code 1
pg_ctl: server does not shut down
........................................................... failed
HINT: The "-m fast" option immediately disconnects sessions rather than
```

Where `omni_httpd` extension is created and then promptly shut down. 

Some of the omni_httpd workers seemingly don't get shut down. I've spent some time on this and couldn't pinpoint it exactly. 

One of the subtle observations I had is that, seemingly, postmaster was sending the termination signals to one of the processes twice and skipping one of the processes. This is my biggest clue so far.

P.S. omni_httpd, one of the oldest extensions, is not a shining example of code quality. It was very much experimental and is due to be replaced by generalized workers (omni_workers) – but for the time being, it'd be cool to solve the edge cases it has!

---

**Bounty size: $250**

[Read more about bounties](https://github.com/omnigres/omnigres/wiki/Bounties)

steve-chavez:
For extra info, this occurred on https://github.com/omnigres/omnigres/actions/runs/9830277033/job/27136953727

yrashk:
Observing this _much_ more rarely.

PR #591, opened 2024-07-12
[temp] failure investigation

yrashk:
See #585 and #584

Trying to see if it will fail without any changes.


yrashk:
[This failure](https://github.com/omnigres/omnigres/actions/runs/9913244689/job/27392282650?pr=591) suggests that the worker termination problem is not caused by changes in #550

PR #592, opened 2024-07-13
feat: add support for dynamically evaluating SQL

ebellani:


yrashk:
Great start! I appreciate the effort. Considering that omni_sql already requires a native extension to be loaded, what do you think about using & extending its capabilities to parse and analyze queries?

ebellani:
> Great start! I appreciate the effort. Considering that omni_sql already requires a native extension to be loaded, what do you think about using & extending its capabilities to parse and analyze queries?

sure thing.

ebellani:
closing to focus on the C version

PR #593, opened 2024-07-18
Problem: cmake setup is not suitable for packaging

steve-chavez:
Currently the cmake setup forces `CPMAddPackage` for various dependencies, it doesn't allow to get them from a package mananger.

This makes it hard to create a Nix package of omnigres, for one. Other Linux distributions will have the same problem.

Solution: Try to get dependencies from the system and if not available use `CPMAddPackage`.

Some dependencies can be found with `find_package` and others with `pkg_check_modules`.

List of dependencies covered:

- [x] pugixml
- [x] uriparser
- [x] h2o
- [x] stc
- [x] metalang99
- [x] libfyaml


extensions/omni_httpc/CMakeLists.txt
```diff
@@ -27,12 +27,14 @@ add_postgresql_extension(
...
+target_link_libraries(omni_httpc ${H2O_LIBRARIES} libpgaug libomnisql libomnitypes ${OPENSSL_LIBRARIES})
 
 target_compile_definitions(omni_httpc PRIVATE H2O_USE_LIBUV=0)
 
-get_target_property(_h2o_deps libh2o INCLUDE_DIRECTORIES)
-target_include_directories(omni_httpc PUBLIC $<TARGET_PROPERTY:libh2o,INCLUDE_DIRECTORIES>)
+if (TARGET libh2o-evloop)
+  get_target_property(_h2o_deps libh2o-evloop INCLUDE_DIRECTORIES)
+  target_include_directories(omni_httpc PUBLIC $<TARGET_PROPERTY:libh2o-evloop,INCLUDE_DIRECTORIES>)
+endif()
```
steve-chavez, line 37: Maybe this can be DRYed by settting the variable inside `FindH2O.cmake`.

steve-chavez:
There's a problem with cmake on MacOS. Using `find_package(uriparser)` and
`FindUriparser.cmake` makes it recurse infinitely.

Tempted to delete `FindUriparser.cmake` to avoid the error since uriparser is only used in a single place.

yrashk:
> There's a problem with cmake on MacOS. Using `find_package(uriparser)` and `FindUriparser.cmake` makes it recurse infinitely.
> 
> https://github.com/omnigres/omnigres/actions/runs/9999050393/job/27639257317?pr=593
> 
> Tempted to delete `FindUriparser.cmake` to avoid the error since uriparser is only used in a single place.

I see you added `find_package(uriparser)` there. That's what [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) does: 

**Module mode**
In this mode, CMake searches for a file called Find<PackageName>.cmake

^^ this recurses

May be the following can be used? In the context of [Nix] packaging, at least.

**Config mode**
In this mode, CMake searches for a file called <lowercasePackageName>-config.cmake or <PackageName>Config.cmake.

yrashk:
@steve-chavez thanks for the initial contribution – it helped to get #642 underway!

Let me know if #642 is helpful for Nix packaging (at least as a first step)

PR #594, opened 2024-07-19
Problem: sharing memory between backends

yrashk:
While we can use DSM/DSA or something similar, it's still painfully difficult: managing named segments, translating pointers, etc.

Wouldn't it be great to have a shared address space, in anticipation of Postgres switching to the thread model?

Solution: add the concept of "omni memory" to omni

This omni memory is allocated in the postmaster and is typically equal to the size of physically available RAM, or `omni.omni_memory_size` configuration parameter.

On Linux, if configured (`sysctl vm.overcommit_memory` set to 1), we can significantly overcommit. On macOS, it seems to work out of the box (I was able to overcommit to the tune of 50TB)

Omni then provides an API to reserve chunks of memory, as well as iterate over them. Currently, chunk release API is a no-op because it's a very simple allocation scheme: a bump allocator. It can't free. This is rather temporary so we don't have to focus on something more complex.


yrashk:
In order to avoid unnecessary on-demand paging, perhaps it would be worth to move page allocation metadata to the front.

It's worth considering making shared memory allocations page-aligned so that we can still mark the overarching chunk as PROT_NONE and then mprotect individual page-aligned chunks however we'd like.

This further illustrates the segmentation between omni memory being the simple memory pages provider and shared memory API being the reference counting, module-aware allocator.

extensions/omni/omni_memory.c
```diff
@@ -0,0 +1,120 @@
...
+  }
+  void *address;
+  size_t page_size = sysconf(_SC_PAGESIZE);
+  do {
+    address =
+        mmap(NULL, omni_memory_size, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    if (address == MAP_FAILED) {
+      size_t new_omni_memory_size = omni_memory_size / 2;
+      if (new_omni_memory_size < page_size) {
+        ereport(WARNING, errmsg("Failed to mmap %lu bytes for omni memory, none will be available",
```
steve-chavez, line 99: How about showing errno `strerror(errno)` in these warnings? I see `mmap` sets it when it fails.

extensions/omni/omni_memory.c
```diff
@@ -0,0 +1,120 @@
...
+                                omni_memory_size, new_omni_memory_size));
+        return;
+      }
+      ereport(WARNING,
+              errmsg("Failed to mmap %lu bytes for omni memory, attempting to allocate %lu",
+                     omni_memory_size, new_omni_memory_size));
+      omni_memory_size = new_omni_memory_size;
+    }
+  } while (address == MAP_FAILED);
```
steve-chavez, line 108: Could this cause an infinite loop somehow? I see mmap is retried with half of the memory size and then quit if its less than the page size. But could there be another reason why it fails?

Perhaps have a max number of retries to be on the safe side?

PR #596, opened 2024-07-31
Problem: having to create API endpoints or use GraphQL (🚀 omni_sql 0.3.0)

yrashk:
Creating API endpoints makes it much more tedious to develop a backend portion of a system, requiring constant interaction between frontend and backend development.

Employing a query language abstraction like GraphQL helps but creates yet another layer of translation.

Solution: make it easier to execute dynamic SQL on the backend

While this is not a complete solution, especially in terms of security – there is a different effort for this, this bit illustrates and paves the way to arbitrary SQL execution. It's mostly implemented in PL/pgSQL for simplicity's sake. It might need to be rewritten in a native extension at a later point.

It's a bit basic right now and passage of parameters is not perfect.

It has one problem: it wouldn't return the number of rows affected. This seems to be an omission in the SPI implementation in Postgres itself and is being investigated separately.

---

cc @ebellani 

Relevant issue: #592


yrashk:
Summary of the potential Postgres bug:

image

ebellani (commented):
:+1:

ebellani (approved):

PR #597, opened 2024-08-02
Problem: mismatching identity types

yrashk:
Sometimes debugging long and winding SQL code is hard on the eyes and the brain, even though mistakes can be trivial!

Something as simple as using a wrong column in a join can cost hours in debugging only to find out that we compare two integers that are not the same.

Many programming languages have dealt with this by having a concept of a "newtype" or wrapping the underlying type.

Postgres has domain types, but they are effectively equivalent to the underlying type when it comes to type comparison, so they don't solve this problem.

Solution: introduce "identity types"

These are implemented as effectively duplicates of underlying integer types. This allows us to distinguish the types.

```sql
create table users
(
   id user_id primary key default user_id_nextval()
);
create table orders
(
   id      order_id primary key default order_id_nextval(),
   user_id user_id not null references users (id)
);

select * from users inner join orders on orders.id = users.id

-- ERROR: operator does not exist: order_id = user_id
-- ^^^ this just saved us precious time
```

In the future, maybe we should consider supporting other types like UUID? omni_seq types?


yrashk:
Intro: https://www.loom.com/share/5aef4295495b4577b436b359246febe4?sid=e126f061-049c-4fb3-a3d4-9777cf29693c

PR #601, opened 2024-08-02
Problem[1]: the whole chain behind omni_sql.execute lacks comments

ebellani:
Problem[2]: omni_sql.execute_parameterized does not use SPI_is_cursor_plan to figure out if rows are returned, making it more complex than needed.

Problem[3]: too many things being done by the the execute_parameterized function.

Solution[1]: Add comments to the main functions created, specially for the pl/pgsql ones that can be consulted by a PostgreSQL user via \df+.

Solution[2]: replace calls to is_returning_statement by SPI_is_cursor_plan.

Solution[3]: Break down the main routine into subroutines, document them.


extensions/omni_sql/src/execute.sql
```diff
@@ -27,4 +27,7 @@ begin
     return;
 end;
 $pgsql$
-    language plpgsql;
\ No newline at end of file
+    language plpgsql;
+
+
+comment on function omni_sql.execute is 'Runs an arbitrary SQL (possibly prepared) statement. If the statement is a prepared statement, one can define more precise types for the statement using the "types" parameter.'
```
yrashk, line 33: I like this idea, but it is important to note that none of the functions across the platform have this. This can be a start.

extensions/omni_sql/omni_sql.c
```diff
@@ -583,19 +583,140 @@ Datum is_returning_statement(PG_FUNCTION_ARGS) {
...
+  int nargs;
+  // the (optional) types of the parameters of the prepared statement
+  Oid *types;
+  // is any parameter null?
+  char *nulls;
+  // the arguments to replace the parameters of the prepared statement
+  Datum *values;
+} ExecCtx;
+
+/*
```
yrashk, line 611: Minor: for Doxygen (which we use in some parts of the project, but not actively), `/**` is the trigger (as opposed to `/*`)

yrashk (approved):
Great effort and good catches!

In the future, would make sense to make changes in a sequence to avoid mixing them together. Would make our lives easier down the road.

Can you add a CHANGELOG, covering the previous release as well (I forgot to do it for the previous release)?

yrashk:
clang-format failing: `make fix-format` will fix it.

extensions/omni_sql/src/execute.sql
yrashk: I am also not sure whether we should talk about prepared statements here as it may be confused with [`PREPARE`](https://www.postgresql.org/docs/current/sql-prepare.html)

ebellani: let me retry the text.

ebellani: > I am also not sure whether we should talk about prepared statements here as it may be confused with [`PREPARE`](https://www.postgresql.org/docs/current/sql-prepare.html)

I was thinking about that. Since we are expecting the end user to heavily interact with the DBMS, accessing the docstrings live would be a help, imho.

yrashk: I agree that this can be of help indeed. The endpoint itself can be used to query the docstrings. Was just pointing out that the rest of the extensions don't have this (yet)

yrashk:
Great job, thank you! 🫶

PR #605, opened 2024-08-08
Problem: impossible to establish presence of omni

yrashk:
This is problematic for some extensions because they can't detect whether omni has been preloaded in `_PG_init`, causing them to be loaded successfully even though they may not be actually operational.

This may shift the need to check whether `_Omni_init` has been called in exported functions, which is adding checks for every invocation and is error-prone (it is easy to forget to do so)

Solution: establish a "marker" that omni is loaded

This is done through a rendezvous variable. Relatively low-key way to do so and it can be then checked in extension's `_PG_init` and if omni is required for operations, it can error out.


yrashk:
Should we include the pointer to base address of the library (what `dlopen` would have returned) in the marker so we can access functions in `omni` in the future, without doing `dlopen`?

This would likely be involving  a call to `dladdr` on omni's side.

PR #606, opened 2024-08-09
Problem: retrying transactions (🚀 omni_txn 0.2.0)

yrashk:
This requires additional logic on the implementation, which is typically a variation of retries and bailing out.

Doing so manually every time is time-consuming and error-prone.

Solution: allow for automatic retries using SQL


yrashk:
Initial prototype: likely not every edge case is thought of. Very primitive retries, no backoff.

extensions/omni_txn/docs/retry.md
```diff
@@ -0,0 +1,50 @@
...
+
+Now, if we have these two simultaneous transactions happening, the second one may have committed first:
+
+```postgresql
+--- (1)
+call omni_txn.retry($$
+select quantity from inventory where product_name = 'Widget';
+update inventory set quantity = quantity + 20
+       where product_name = 'Widget'
+$$);
```
steve-chavez, line 42: Seems useful! Have you thought of a global configuration? When enabling serializable isolation is usually cumbersome to retry individual statements, it'd be more convenient to have a GUC.. with that it can be enabled per session, role or db.

ebellani:
How about adding backoff jittering?

```c
static float8 backoff (float8 cap, float8 base, int32 attempt){
  float8 exp = (float8) (1 << attempt);
  return Min(cap, base * exp);
}

/**
 * See https://stackoverflow.com/questions/13408990/how-to-generate-random-float-number-in-c#comment56659626_13409133
 */
static float random_between(float8 a) {
  return (float)((double)rand()/(double)(RAND_MAX/a));
}

static float8 sleep_for(float8 cap, float8 base, int32 attempt){
  return random_between(backoff(cap, base, attempt));
}
```

then call it with something like 

```c
    DirectFunctionCall1(pg_sleep, Float8GetDatum(sleep_for(0.0001, 0.000001, retry_attempts)));
```

yrashk:
@ebellani I think it's a good idea – and calling pg_sleep is fairly important here, I think, as it will ensure we're checking for interrupts and such.

I am also thinking about limiting the retries by max time, too.

ebellani:
> @ebellani I think it's a good idea – and calling pg_sleep is fairly important here, I think, as it will ensure we're checking for interrupts and such.
> 
> I am also thinking about limiting the retries by max time, too.

I think if you add a max number of attempts it already caps the max time if you use the backoff fn (cap*max_attemp, basically)

yrashk:
> I think if you add a max number of attempts it already caps the max time if you use the backoff fn (cap*max_attemp, basically)

We already have max number of attempts.

extensions/omni_txn/omni_txn.c
```diff
@@ -0,0 +1,81 @@
...
+
+static int32 retry_attempts = 0;
+
+PG_FUNCTION_INFO_V1(retry);
+
+Datum retry(PG_FUNCTION_ARGS) {
+  if (PG_ARGISNULL(0)) {
+    ereport(ERROR, errmsg("transaction statements argument is required"));
+  }
+  if (IsTransactionBlock()) {
```
yrashk, line 23: There is still a case where it doesn't work: 

When we're not connected over a backend going through the wire protocol, but instead using something like a background worker, this check may not be doing anything.

We may very well be in an ongoing transaction but `IsTransactionBlock()` will return `false`.

So far I found that starting an implicit transaction (`BeginImplicitTransactionBlock`) around the execution makes this check effective, but it also makes it impossible to use this function even if we just wanted to simply call the procedure.

I simulated the change here:

```diff
diff --git a/extensions/omni_httpd/http_worker.c b/extensions/omni_httpd/http_worker.c
index ca473f1..de46842 100644
--- a/extensions/omni_httpd/http_worker.c
+++ b/extensions/omni_httpd/http_worker.c
@@ -652,6 +652,7 @@ static int handler(request_message_t *msg) {
   SetCurrentStatementStartTimestamp();
   StartTransactionCommand();
   PushActiveSnapshot(GetTransactionSnapshot());
+  BeginImplicitTransactionBlock();
   SPI_connect();
 
   bool succeeded = false;
@@ -894,6 +895,7 @@ static int handler(request_message_t *msg) {
 
 cleanup:
   SPI_finish();
+  EndImplicitTransactionBlock();
   PopActiveSnapshot();
   if (succeeded) {
     CommitTransactionCommand();
```

This could be better. Thinking of a better method to ensure that the transaction has _actually_ started.  🤔 


🤔

yrashk: A bigger problem here is that I don't yet know how to make this work in `omni_httpd` – currently, all handling goes through a transaction. It does work, but erroneously, because we don't begin a transaction block.

* Should omni_httpd (and future workers) handlers be configurable as retriable transactions?
* Should we support multi-transactional workloads?

yrashk: Upcoming omni_httpd supports multi-transactional handlers #550

PR #609, opened 2024-08-14
Problem: omni_httpd.handlers approach

yrashk:
It limits us as to what would be possible: we always need to execute a
query to serve a request.

So, for example, we won't be able to call a non-atomic procedure, even
if we wanted to.

Solution: switch to direct function call

omni_httpd now exposes a `handler` function that must remain in place
but can be replaced. This function will serve any and all requests.

I am occassionally observing a slightly better RPS with this implementation.

---

What's missing:

- [ ] Documentation changing away from the `handlers` table to the `handler` function
- [ ] Tests adopting `handler` function as well (we can keep a test that tests the transitional `handlers`->`handler` generator trigger)


yrashk:
While the most primitive case of transaction isolation level change works, something more complicated wouldn't:

```sql
 create or replace function omni_httpd.handler(int, omni_httpd.http_request) returns omni_httpd.http_outcome
        language plpgsql as
    $$
    begin
        if $2.path = '/iso' then
            set transaction isolation level serializable;
        end if;
        return omni_httpd.http_response(current_setting('transaction_isolation'));
    end;
    $$
```

This is because evaluating `$2.path = /iso` goes through `exec_eval_expr`, which will acquire and set the snapshot, preventing setting transaction isolation level (it checks whether `FirstSnapshotSet` is true)

yrashk:
> While the most primitive case of transaction isolation level change works, something more complicated wouldn't:
> 
> ```sql
>  create or replace function omni_httpd.handler(int, omni_httpd.http_request) returns omni_httpd.http_outcome
>         language plpgsql as
>     $$
>     begin
>         if $2.path = '/iso' then
>             set transaction isolation level serializable;
>         end if;
>         return omni_httpd.http_response(current_setting('transaction_isolation'));
>     end;
>     $$
> ```
> 
> This is because evaluating `$2.path = /iso` goes through `exec_eval_expr`, which will acquire and set the snapshot, preventing setting transaction isolation level (it checks whether `FirstSnapshotSet` is true)

But it does work when done this way:

```sql
    create or replace function omni_httpd.handler(int, omni_httpd.http_request) returns omni_httpd.http_outcome
        language plpgsql as
    $$
    begin
        if $2.path = '/iso' then
            rollback;
            set transaction isolation level serializable;
        end if;
        return omni_httpd.http_response(current_setting('transaction_isolation'));
    end;
    $$
```

PR #612, opened 2024-08-15
Problem: FlyCI macOS runners shutting down next month

yrashk:
See the announcement:
https://flyci.net/blohttpsg/flyci-discontinue-macos-runners

Solution: switch to WarpBuild's runners


yrashk:
No difference between this and #611

PR #628, opened 2024-08-21
Problem: omni_httpd does not support WebSockets

yrashk:
This is a major problem for anything remotely interesting these days.

Solution: devise the first version of WebSocket handling

Current approach definitely has deficiencies. For example, using UNIX domain sockets can be problematic here. It's good that sending a message to UNIX domain socket is directly connecting us with the consumer (the networking thread) thus ensuring that we are not simply buffering the messages. But if, for some reason, the consumer is not reading, it'll block us.

An even bigger problem is that we currently won't try to send messages to any other worker and will simply schedule messages for the current buffer. It's okay for the initial prototype, but won't scale well.

### DIFF
--- a/cmake/FindH2O.cmake
+++ b/cmake/FindH2O.cmake
@@ -4,7 +4,14 @@ include(CheckCSourceCompiles)
 set(_h2o_options)
 list(APPEND _h2o_options "WITH_MRUBY OFF")
 list(APPEND _h2o_options "DISABLE_LIBUV ON")
+# This is to make h2o find WSLAY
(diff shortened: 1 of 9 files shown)


ebellani:
is it possible to break this up in smaller bits? Too large of a PR. 

Anyway, LGTM

yrashk:
Kind of hard to do with interconnected pieces.

yrashk:
This is truly an early/experimental version. A lot of things will change and improve. There are no tests at the moment as the WebSocket client is genuinely experimental and unpublished. However, I think we need to merge this to keep moving forward.

PR #633, opened 2024-08-22
Fix Omni Seq's include of PostgreSQL 16's server header

### LINKED ISSUE #631: The current build fails to pick up PGDG PostgreSQL server include files on Ubuntu Noble
The build doesn't account for the fact that, on some systems (E.G. Ubuntu Noble LTS), server development files are packaged separately from client development files.

image

image

dysinger:
Problem:

Compiling on Ubuntu Noble LTS against PGDG PostgreSQL 16 is broken.

Solution:

FindPostgreSQL cmake file already includes both client & server include directories. All we have to do is include the header we want without being specific about the path. This works with both RedHat and on Ubuntu PGDG packages. Server headers are a separate package on Ubuntu.

Closes #631

### DIFF
--- a/extensions/omni_seq/id.h
+++ b/extensions/omni_seq/id.h
@@ -34,7 +34,7 @@
 #include <utils/builtins.h>
 #if PG_MAJORVERSION_NUM >= 16
 #include "port/pg_bswap.h"
-#include "server/varatt.h"
+#include <varatt.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


dysinger:
I see some CI checks didn't pass but it's just wanting a version bump for this.

yrashk:
> I see some CI checks didn't pass but it's just wanting a version bump for this.

Yes, merges to master have to bump affected components. Will see to it. 

Thank you for the fix!

yrashk:
Thank you! Superseded by #637

PR #634, opened 2024-08-22
Problem: `regproc[edure]` have limited usefulness

yrashk:
Sure, they denote that an OID is a function/procedure, and, in case of `regprocedure`, that function has certain arguments.

But the type itself does not convey the signature, which makes it not possible for us to call these function.

Solution: design "function signature types"

This type is also an OID but we also store information about its full signature and we validate on casts that the functions match the entire signature.

These makes these types represent the actual function signature, making them callable. One can compose them into interfaces through composite types.

---

Missing:

- [x] tests
- [x] documentation
- [x] extconfig (make sure the types will be dumped)

### DIFF
--- a/extensions/omni_types/migrate/3_function_signature_type.sql
+++ b/extensions/omni_types/migrate/3_function_signature_type.sql
@@ -1,0 +1,12 @@
+create table function_signature_types
+(
+    typ       regtype primary key unique,
+    arguments regtype[] not null,
+    return    regtype   not null
+);
+
+select pg_catalog.pg_extension_config_dump('function_signature_types', '');
+
+/*{% include "../src/function_signature_type_of.sql" %}*/
+
+/*{% include "../src/function_signature_type.sql" %}*/
--- a/extensions/omni/test/tests/upgrade_extra_deps.yml
+++ b/extensions/omni/test/tests/upgrade_extra_deps.yml
@@ -10,7 +10,7 @@ instance:
   - create extension omni_test version '1' cascade
   - create extension omni
   # We use omni_types to test how dependencies are upgraded
-  - create extension omni_types
+  - create extension omni_types cascade
 
 tests:
 
--- a/extensions/omni_types/CHANGELOG.md
+++ b/extensions/omni_types/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [0.2.0] - TBD
 
+### Added
+
+* Function signature types [#634](https://github.com/omnigres/omnigres/pull/634])
+
 ### Fixed
 
 * Sum types will be now included into `pg_dump`-produced dumps [#636](https://github.com/omnigres/omnigres/pull/636])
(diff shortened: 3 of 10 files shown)


extensions/omni_types/migrate/3_function_signature_type.sql
```diff
@@ -0,0 +1,110 @@
+create table function_signature_types
+(
+    typ       regtype primary key unique,
+    arguments regtype[] not null,
```
yrashk, line 4: maybe use `oidvector` here?

extensions/omni_types/migrate/3_function_signature_type.sql
```diff
@@ -0,0 +1,110 @@
...
+    ret_type := signature[cardinality(signature)];
+    insert into omni_types.function_signature_types (typ, arguments, return)
+    values (name::regtype, trim_array(signature, 1), ret_type);
+
+    -- Define casts in and out of the type
+    execute format('create cast (regtype as %I) without function', name);
+    execute format('create cast (%I as regtype) without function', name);
+
+    -- Define caller function
+    execute format('create function %I(%I %s) returns %s language c as %L, %L', 'call_' || name, name,
```
yrashk, line 102: I don't love `call_NAME` and would have preferred `NAME`, but for zero-arity functions, this conflicts with the syntax for casts.

yrashk: Also, can't do `[schema.]call` because it won't work for functions with different return types.

yrashk:
It would be great to be able to filter procedures by their conforms to a signature (without erroring out) 

cc @thenonameguy

yrashk:
It's also important to add namespace support to function casting and conformance testing

yrashk:
@thenonameguy 

```sql
with conforming as (select t2i_conforming_function((nspname || '.' || proname)) as func from pg_proc 
     inner join pg_namespace on pg_namespace.oid = pronamespace) 
select func, call_t2i(func, 'test') from conforming where func is not null;
          func          |  call_t2i  
------------------------+------------
 textlen(text)          |          4
 hashtext(text)         | 1771415073
 octet_length(text)     |          4
 length(text)           |          4
 length(text)           |          4
 character_length(text) |          4
 character_length(text) |          4
 char_length(text)      |          4
 octet_length(text)     |          4
 octet_length(text)     |          4
 char_length(text)      |          4
 length(text)           |          4
 length(text)           |          4
 ascii(text)            |        116
 length(text)           |          4
 length(text)           |          4
 octet_length(text)     |          4
 length(text)           |          4

```

yrashk:
@thenonameguy 

Slightly better query:

```sql
select conforming from pg_proc inner join pg_namespace on pg_namespace.oid = pronamespace inner join lateral (values (t2i_conforming_function(nspname || '.' || proname))) t(conforming) on conforming is not null;
```

thenonameguy:
> @thenonameguy
> 
> Slightly better query:
> 
> ```sql
> select conforming from pg_proc inner join pg_namespace on pg_namespace.oid = pronamespace inner join lateral (values (t2i_conforming_function(nspname || '.' || proname))) t(conforming) on conforming is not null;
> ```

Shouldn't the `t2i_conforming_function(nspname || '.' || proname)` be:
`quote_ident(nspname) || '.' || quote_ident(proname)`

PR #637, opened 2024-08-25
Problem: failing to compile omni_seq in some cases

yrashk:
In particular, compiling on Ubuntu Noble LTS against PGDG PostgreSQL 16 is failing with an error (not being able to find a header file)

Solution: adjust referenced header location

FindPostgreSQL cmake file already includes both client & server include directories. All we have to do is include the header we want without being specific about the path. This works with both RedHat and on Ubuntu PGDG packages. Server headers are a separate package on Ubuntu.


yrashk:
Supersedes #633

PR #629, opened 2024-08-21
Add a script and a dockerfile to build RPM pkgs for amd64 & arm64

dysinger:
This commit adds the ability to easily produce RPMs for Omnigres on any capable machine with Docker installed. Please review the ./pkgs/mkpkgs script for notes on setup. Then just call ./pkgs/mkpkgs at the command line. If you only want one archiecture then call `ARCH=linux/arm64 ./pkgs/mkpkgs`.

Please provide as much criticism as you like. I'll fixup commits and rebase.

The build machine setup for this has been exercising this script. Try it yourself if you like.

PR #639, opened 2024-08-26
 Add a script and a dockerfile to build Ubuntu pkgs for amd64 & arm64

dysinger:
This commit adds the ability to easily produce debian packages on any capable
machine with Docker installed. Please review the ./pkgs/mkpkgs script for notes
on setup. Then just call ./pkgs/mkpkgs at the commandline. If you only want one
archiecture then call `ARCH=linux/arm64 ./pkgs/mkpkgs`.


dysinger:
Needs to be rebased on #629 after it's merged

dysinger:
image

PR #651, opened 2024-09-27
Problem: Postgres 16 is not the latest version anymore

yrashk:
We need to be able to keep up with the times.

Solution: upgrade to 17 as the latest version

I haven't switched Dockerfile to 17.0 yet as there are no images for arm64 yet (https://github.com/docker-library/postgres/issues/1272)


yrashk:
This won't pick up v17 build matrix, abandoning.

PR #654, opened 2024-10-02
Problem: Postgres regex support of named capture groups (🚀 omni_regex 0.1.0)

yrashk:
It doesn't support these! However, I'd really love to have it supported. For example, it'll be very useful for extracting named captures for routing regular expressions in HTTP handling.

Solution: release omni_regex based on PCRE2

It is based on [pgpcre](https://github.com/petere/pgpcre) by @petere with numerous improvements.

---

What's missing?

- [x] Documentation
- [ ] Better commentary in the source code


ebellani:
:+1:

PR #663, opened 2024-10-17
Problem: omni_txn.retry is rather inflexible

yrashk:
We can only pass a string as a query, and that's it.

Solution: allow the query to be parameterized

This will allow preparing data separately from the query.

---

cc @ebellani


ebellani (commented):
:+1:

PR #674, opened 2024-10-23
Add timeout parameter to omni_txn.retry function to specify transaction duration

yadavshubham01:
Problem:
The current implementation of the omni_txn.retry function uses a max_attempts parameter to control how many times a transaction should be retried. However, this approach does not allow specifying how long the system should continue retrying, which makes it difficult to control the overall retry duration. As a result, transactions may attempt retries beyond a practical time limit, potentially leading to longer-than-desired delays or system timeouts.

Solution:
The omni_txn.retry function has been updated to introduce a timeout parameter, which specifies the maximum duration (in milliseconds) for which retry attempts should be made. This allows for more fine-grained control over retry behavior by setting a time limit for retry attempts, rather than relying solely on the number of attempts.

Additionally, the following updates have been made:

-Test cases in retry.yml have been updated to include scenarios that validate the timeout functionality.
-Documentation has been added to explain the usage of the timeout parameter, ensuring clarity for future developers.
-The function implementation has been refactored to ensure the new parameter integrates smoothly with the existing logic while maintaining readability.

This solution provides a more flexible way to handle transaction retries by focusing on time-based limits instead of only retry counts, thus improving the practical applicability of the retry mechanism.


extensions/omni_httpd/src/routing/make_router.sql
```diff
@@ -0,0 +1,62 @@
+create or replace function make_router(
```
yrashk, line 1: This file does not belong here.

extensions/omni_txn/omni_txn.c
```diff
@@ -114,6 +114,11 @@ Datum retry(PG_FUNCTION_ARGS) {
     collect_backoff_values = PG_GETARG_BOOL(3);
   }
 
+   // Timeout logic
+  int64 timeout_microsecs = 30000000; // Example: 30 seconds timeout
```
yrashk, line 118: Hard-coded value, not retrieved from the arguments?

extensions/omni_txn/migrate/2_retry.sql
```diff
@@ -1,10 +1,12 @@
 create procedure retry(stmts text, max_attempts int default 10, repeatable_read boolean default false,
-                       collect_backoff_values boolean default false)
+                       collect_backoff_values boolean default false , timeout int default 0 )
```
yrashk, line 2: Tests use `interval` but here is an int?

yrashk (changes requested):
Thank you for submitting this PR. It appears it is incomplete? I've also left some in-band notes.

Please convert it to a Draft PR if it is not ready!

extensions/omni_httpd/src/routing/make_router.sql
yadavshubham01: sorry of this i removed it

extensions/omni_txn/omni_txn.c
```diff
@@ -114,6 +114,14 @@ Datum retry(PG_FUNCTION_ARGS) {
     collect_backoff_values = PG_GETARG_BOOL(3);
   }
 
+   // Timeout logic
+  int64 timeout_microsecs = 30000000; // Default timeout value (30 seconds)
+  if (!PG_ARGISNULL(5)) {
+  timeout_microsecs = PG_GETARG_INT64(5); // Retrieve timeout from the argument
```
yrashk, line 120: There is `PG_GETARG_INTERVAL_P` – interval is definitely not a 64-bit int but a pointer to [Interval](https://doxygen.postgresql.org/structInterval.html)

extensions/omni_txn/omni_txn.c
```diff
@@ -210,6 +218,14 @@ Datum retry(PG_FUNCTION_ARGS) {
             // go back to the context where we were
             MemoryContextSwitchTo(current_mcxt);
           }
+           // Check for timeout before retrying
+           TimestampTz current_time = GetCurrentTimestamp();
+             if (TimestampDifferenceExceeds(start_time, current_time, timeout_microsecs)) {
```
yrashk, line 223: I believe the function is not forward-declared but declared below. Would it make sense to move the function definition up?

extensions/omni_txn/tests/retry.yml
```diff
@@ -117,14 +117,14 @@ tests:
     - count: 1
       iso_level: serializable
 
-- name: max attempts
+- name: max attempts with timeout
```
yrashk, line 120: We want to test all relevant behaviors – max_attempts, timeouts, etc. – whether they follow the logic we decide on

yrashk (changes requested):
Thank you for updating the PR, please find my comments attached.

extensions/omni_txn/tests/retry.yml
yrashk: This likely means separate test for each case.

extensions/omni_txn/migrate/2_retry.sql
```diff
@@ -1,10 +1,12 @@
-create procedure retry(stmts text, max_attempts int default 10, repeatable_read boolean default false,
-                       collect_backoff_values boolean default false)
+create procedure retry(stmts text, max_attempts int default null,, repeatable_read boolean default false,
```
yrashk, line 1: extra comma after max_attempts

extensions/omni_txn/migrate/2_retry.sql
```diff
@@ -1,10 +1,12 @@
...
+                       collect_backoff_values boolean default false ,timeout interval default null )
```
yrashk, line 2: no leading space before timeout

extensions/omni_txn/migrate/2_retry.sql
```diff
@@ -1,10 +1,12 @@
...
     language c as
 'MODULE_PATHNAME';
 
 comment on procedure retry is $$
-Retry serializable transaction on statements `stmts`, `max_attempt` number of times (10 by default). `collect_backoff_values` controls if the backoff values used for sleeping will be recorded for debugging/testing purposes (false in orded to increase performance)
+Retry serializable transaction on statements `stmts`, `max_attempt` number of times (10 by default). 
+`collect_backoff_values` controls if the backoff values used for sleeping will be recorded for debugging/testing purposes (false in order to increase performance).
+`timeout` specifies the maximum time in seconds to attempt the transaction.
```
yrashk, line 9: We still need to specify expected behaviors:

* What happens if both timeout and max attempts are specified?
* Do defaults apply if one of the other values is specified?

yrashk:
Errors in CI:

```
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:120:33: warning: implicit declaration of function ‘PG_GETARG_INTERVAL_P’; did you mean ‘PG_GETARG_BYTEA_P’? [-Wimplicit-function-declaration]
  120 |    Interval *timeout_interval = PG_GETARG_INTERVAL_P(5); // Retrieve timeout from the argument
      |                                 ^~~~~~~~~~~~~~~~~~~~
      |                                 PG_GETARG_BYTEA_P
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:120:33: warning: initialization of ‘Interval *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
[ 91%] Linking C shared module omni_os--0.1.1.so
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:126:28: warning: implicit declaration of function ‘GetCurrentTimestamp’ [-Wimplicit-function-declaration]
[ 93%] Linking C shared module no_fun--1.so
  126 |   TimestampTz start_time = GetCurrentTimestamp(); // Get start time
      |                            ^~~~~~~~~~~~~~~~~~~
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:185:8: error: redefinition of ‘retry’
  185 |   bool retry = true;
      |        ^~~~~
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:127:8: note: previous definition of ‘retry’ with type ‘_Bool’
  127 |   bool retry = true;
      |        ^~~~~
[ 93%] Linking C shared module omni_web--0.1.1.so
[ 93%] Built target no_fun
[ 93%] Built target omni_os
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c: In function ‘TimestampDifferenceExceeds’:
/home/runner/work/omnigres/omnigres/extensions/omni_txn/omni_txn.c:354:5: warning: implicit declaration of function ‘TimestampDifference’; did you mean ‘list_difference’? [-Wimplicit-function-declaration]
[ 93%] Built target omni_edbg
  354 |     TimestampDifference(start, current, &elapsed);
      |     ^~~~~~~~~~~~~~~~~~~
      |     list_difference
[ 93%] Built target omni_var
```

yadavshubham01:
just fix some error now you can review it

yadavshubham01:
@yrashk i will fix that error can you re-run workflows

yadavshubham01:
@yrashk can you please saw why i am get ci error in test

yrashk:
The CI error is happening because you modified a dependency in-band (2_retry) and another [migration](https://github.com/omnigres/omnigres/blob/master/extensions/omni_txn/migrate/3_retry_params.sql) is trying to drop the previously known signature of `retry`.

Instead, you should add a new, incremented migration and drop and create a new signature, similar to 3_retry_params (but accounting for the fact that 3's signature is already different)

This is done so that the extensions are upgradable.

yadavshubham01:
@yrashk i just add new file in migration 5_retry_udpate can please review it

extensions/omni_txn/migrate/2_retry.sql
```diff
@@ -1,10 +1,12 @@
-create procedure retry(stmts text, max_attempts int default 10, repeatable_read boolean default false,
-                       collect_backoff_values boolean default false)
+create procedure retry(stmts text, max_attempts int default null, repeatable_read boolean default false,
```
yrashk, line 1: This migration should not be changed.

extensions/omni_txn/migrate/5_retry_update.sql
```diff
@@ -0,0 +1,21 @@
+drop procedure if exists retry(stmts text, max_attempts int, repeatable_read boolean,
+                               collect_backoff_values boolean, params record);
+
+
+create procedure retry(
+    stmts text, 
+    max_attempts int default 10, 
+    repeatable_read boolean default false, 
+    collect_backoff_values boolean default false, 
+    timeout interval default null,          -- Adding timeout parameter or other changes
```
yrashk, line 10: It is better to add new parameters at the end as a general practice to ensure that those using these functions without names, don't suddenly get a signature mismatch.

extensions/omni_txn/migrate/5_retry_update.sql
```diff
@@ -0,0 +1,21 @@
...
+    params record default null::record
+)
+    language c as
+'MODULE_PATHNAME';
+
+
+comment on procedure retry is $$
+Retry serializable transaction on statements `stmts`, retrying `max_attempt` number of times (10 by default).
+`collect_backoff_values` controls whether the backoff values used for sleeping will be recorded for debugging/testing purposes.
+You can also specify a `timeout` to control the total time the retry attempts will take.
```
yrashk, line 20: We need to specify the interaction between timeout and max_attempts, how do they work in various scenarios.

yadavshubham01:
hey @yrashk please review i think we can fix it

yadavshubham01:
Hey @yrashk can please run workflow I think all test case is passed only one left I try  fix issue let see

yadavshubham01:
Hey @yrashk Is there still an error in this PR

yrashk:
> Hey @yrashk Is there still an error in this PR

Do you run tests locally before pushing?

yadavshubham01:
Ok I got it why I getting this error try fix wait

yadavshubham01:
hey @yrashk sorry for mistake now you can run workflow

yadavshubham01:
i try to solve this error in tests any suggestions??

PR #682, opened 2024-11-01
Problem: limited Docker API coverage

yrashk:
Solution: enable executing in existing containers

This can be useful for container inspection


yrashk:
Superseded by #685

PR #687, opened 2024-11-08
Add timeout parameter to omni_txn.retry procedure 

yadavshubham01:
Problem:
The current implementation of the omni_txn.retry function uses a max_attempts parameter to control how many times a transaction should be retried. However, this approach does not allow specifying how long the system should continue retrying, which makes it difficult to control the overall retry duration. As a result, transactions may attempt retries beyond a practical time limit, potentially leading to longer-than-desired delays or system timeouts.

Solution:
The omni_txn.retry function has been updated to introduce a timeout parameter, which specifies the maximum duration (in milliseconds) for which retry attempts should be made. This allows for more fine-grained control over retry behavior by setting a time limit for retry attempts, rather than relying solely on the number of attempts.

Additionally, the following updates have been made:

-Test cases in retry.yml have been updated to include scenarios that validate the timeout functionality.
-The function implementation has been refactored to ensure the new parameter integrates smoothly with the existing logic while maintaining readability.

This solution provides a more flexible way to handle transaction retries by focusing on time-based limits instead of only retry counts, thus improving the practical applicability of the retry mechanism.

### DIFF
--- a/extensions/omni_txn/migrate/2_retry.sql
+++ b/extensions/omni_txn/migrate/2_retry.sql
@@ -4,7 +4,7 @@ create procedure retry(stmts text, max_attempts int default 10, repeatable_read 
 'MODULE_PATHNAME';
 
 comment on procedure retry is $$
-Retry serializable transaction on statements `stmts`, `max_attempt` number of times (10 by default). `collect_backoff_values` controls if the backoff values used for sleeping will be recorded for debugging/testing purposes (false in orded to increase performance)
+Retry serializable transaction on statements `stmts`, `max_attempt` number of times (10 by default). `collect_backoff_values` controls if the backoff values used for sleeping will be recorded for debugging/testing purposes (false in orded to increase performance). `timeout` specifies the maximum duration in seconds for which to attempt the retries (60 seconds by default).
 $$;
 
 create function current_retry_attempt()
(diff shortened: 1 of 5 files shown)


yadavshubham01:
hey @yrashk please this PR

yrashk:
Thank you for sending the PR. It doesn't look like it compiles at the moment, though.

yadavshubham01:
Hi @yrashk ,

Thank you for reviewing my pull request and for bringing the compilation issue to my attention. I appreciate your feedback.

I will investigate the compilation errors and make the necessary adjustments to ensure that the code compiles successfully. Specifically, I will look into the conflicting function signature for TimestampDifferenceExceeds and any other issues that may arise.

Once I have resolved the compilation issues, I will update the PR accordingly. Thank you for your patience, and I will keep you posted on my progress.

yadavshubham01:
hey @yrashk why i am getting this error ??
Screenshot 2024-11-12 183359

yrashk:
> hey @yrashk why i am getting this error ?? Screenshot 2024-11-12 183359

Chances are this is because you are not dropping the latest version of `retry` (this is masked by `if exists` in droppage) and then when you're commenting on it, Postgres can't choose a specific version to comment on because there's more than one.

PR #713, opened 2024-12-14
Problem: omni_httpd needs improvements (🚀 omni_httpd 0.3.0, omni_rest 0.1.1)

yrashk:
Solution: kickstart 0.3.0 development cycle

---

Things to accomplish (this or next release):

- [ ] WebSocket API as procedures
- [ ] Commit before sending responses (with an optional override of the behavior, perhaps?)
- [ ] Specifying middleware without replacing procedure? The most challenging part is that middlewares want their own parameters; also ordering – but that's easier.
- [ ] omni_httpd as a templated extension (would require a patch to omni to bring back pg_proc scanning in addition to pg_extension)


yrashk:
I am still not very happy about having to replace the handler. Not only this is awkward and destructive, it is subject to some [subtle behavior](https://github.com/omnigres/omnigres/pull/750#issuecomment-2600800071)

We need to find something. One thought I just had was, what if we implement a PL that's similar to [PL/Proxy](https://github.com/plproxy/plproxy) but would serve as a "pointer" to a function we want to use as handler? 

Now, let's rewind and see why this current design has pre-set callbacks. The point of it was that this way, we don't need to go into our complex "configuration reloading" scenario (which will improve with I/O and workers extension in the future) because we don't yet have proper custo "cache invalidation." So we kind of lock in onto a single procedure and keep using it.  

It has problems. It must not be deleted otherwise the whole thing will just error out.

So, what would I do in the ideal scenario? Find the procedure(s) on the flight when the cache is invalidated (would require event triggers?)

yrashk:
Continuing the exploration of setting up the handling of requests. What if instead of concerning users with setting up procedures, we had handling defined somewhere in the omni_vfs filesystem? This is a bit different in a sense that it brings a new dependency in. But this is where they _will_ be doing editing of their middleware and routing. At least unless we provide them with a structural way to manage those.

yrashk:
A variation on the current scheme:

What if we scanned for all procedures matching `[int,] request, outcome` and called all of them?  This way, replacement of the extension-owned procedure is not needed.

It would have been nice if we could have specified `set omni_httpd.priority` on them at the time of definition, but this would prevent us from using these for non-atomic scenarios. But we can use other methods to configure that. Comments? A table mapping regprocedure to such parameters?

One interesting challenge here is that if make `int` (listener ID) optional, we'll automatically pick up all such installed functions. Is this good? Bad?

PR #715, opened 2024-12-15
Problem: omni_httpd.handler is a function

yrashk:
However, we're using it as a procedure, enabling non-atomic operations. This is an odd hack.

Solution: make it a proper procedure.


yrashk:
Should we convert websocket functions into procedures, too? This way they can reap the benefits of being multi-transactional.

PR #747, opened 2025-01-14
Problem: object re-creation without impacting dependencies (🚀 omni_sql 0.4.0)

diogob:
In order to validate that statements can be re-executed without any impact to their dependencies we have to ensure they have the OR REPLACE clause. This is necessary for our upcoming migration system. How to identify a statement with OR REPLACE clause?

Solution: Add a function to omni_sql that, given a statement, identifies whether or not it has the OR REPLACE clause.


extensions/omni_sql/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2025-01-14
+
+### Added
+
+* `is_replace_statement` funtion to detect whether a statement has a `OR REPLACE` clause [#747](https://github.com/omnigres/omnigres/pull/747)
```
yrashk, line 12: "function" got misspelled.

an OR REPLACE?

PR #750, opened 2025-01-15
Problem: keeping replaceable objects in sync with source files

diogob:
## Problem
Keeping database functions, views and other replaceable objects in sync with source files while having migrations that have to be executed only once.

This is compounded by the fact that migrations may introduce dependencies for objects in a src directory and vice-versa.

## Solution

Create 2 distinct sources of database objects:

* Commands inside a *migrations* directory are always executed in order and only once.
* Commands inside any other directory (e.g. src) will be interspersed with migrations according to their dependencies and are always executed, but require the statements to have an *OR REPLACE* clause.

We use assemble_schema to figure out dependencies in a temporary database and is_replace_statement to validate the non-migration sources.

The decision to split the functionality into a function to plan and another one to execute is to make it easier to understand, test and in the future implement dry-run or other plan manipulation functionalities.

## TODO
- [x] Handle migration errors.
- [x] How to handle files removed from `src` directories? Should we keep track of these objects and remove them as in `load_from_fs` ?
- [x] The code as is here create several record in `omni_schema.migrations` with the same `name` value. Should we aggregate statements per file? - we will leave the duplicate names for now and add some hashing to identify individual statements with ease.
- [x] Create a test case for cases such as python `requirements.txt`. But it seems that this will fail since it has no `OR REPLACE` clause. Should we create an exception for things processed by `omni_schema.auxiliary_tools` ?
- [x] Docs


yrashk:
> How to handle files removed from src directories? Should we keep track of these objects and remove them as in load_from_fs ?

Gut reaction – yes. But what does add complexity-wise? Should it be part of incremental recording?

diogob:
> > How to handle files removed from src directories? Should we keep track of these objects and remove them as in load_from_fs ?
> 
> Gut reaction – yes. But what does add complexity-wise? Should it be part of incremental recording?

It introduces potential issues where removing a file either breaks the migration system because of unmet dependencies or we drop cascade and potentially create more harm. 
I thought a bit more about this, and leaving the removals to the user discretion on the migrations seems the safest way to proceed with haste and it gets the job done for now.

extensions/omni_schema/src/make_plan.sql
```diff
@@ -0,0 +1,129 @@
+create or replace function make_plan(path text)
```
yrashk, line 1: The code here assumes `local_fs`, however the intention is that we pass VFS instances to make sure it works off any source. You can see it in other places like `assemble_schema`.

`omni_vfs` interface is still a bit clunky so it is understandably a bit more difficult to work with it when it comes to remote connections. This is why `assemble_schema` copies from a local vfs into remote table_fs.

I wonder how difficult will it be for us to make it take fs at this point before rolling out the next release.

Longer-term we just need a good solution to cross-connection FS. Perhaps it might be not too difficult to have a `remote_fs` that goes over fdw?

extensions/omni_schema/docs/reference.md
```diff
@@ -160,4 +160,50 @@ The support for languages is configurable through `omni_schema.languages` table:
...
+* `CREATE OR REPLACE TRANSFORM`
+* `CREATE OR REPLACE AGGREGATE`
+* `CREATE OR REPLACE PROCEDURE`
+* `CREATE POLICY`
+
+This implies that **any object removal has to be done in a migration**, since `make` will not drop any objects even when files in `src` have been removed.
+
+It is also important to note that a temporary database is created in the current cluster (and using the same credentials as the caller) in order to plan the statements to be used by `make`.
+Therefore the user executing this should have all appropriate system privileges.
+Create a blank temporary database to run assemble_schema.
```
yrashk, line 204: I am not sure I understand how the last line is related to `make`.

yrashk (commented):
Here's the first review pass. Great work, see some commentary!

extensions/omni_schema/src/make_plan.sql
yrashk: https://github.com/omnigres/omnigres/pull/755

extensions/omni_schema/src/make_plan.sql
```diff
@@ -0,0 +1,129 @@
...
+    returns table
+            (
+                migration_filename  text,
+                migration_statement omni_sql.statement
+            )
+    language plpgsql
+as
+$$
+declare
+    local_conn_info text := 'host=localhost port=' || (select setting from pg_settings where name = 'port') || ' user=' || current_user;
```
yrashk, line 11: In other parts (omni_test) I've used unix port instead because I suspect it is even more commonly available in local setups. Probably not a big deal, but just as a note:

https://github.com/omnigres/omnigres/blob/master/extensions/omni_test/src/instantiate.sql#L66-L68

extensions/omni_schema/src/make_plan.sql
```diff
@@ -0,0 +1,129 @@
...
+    loop
+            if exists (select from pg_extension where extname = extension_record.extension) then
+                perform dblink(plan_db_conn,
+                               format('create extension if not exists %s cascade', extension_record.extension));
+            end if;
+    end loop;
+
+    for migration in select * from dblink(
+        plan_db_conn,
+        format('select migration_filename, executed_statement, migration_statement, execution_error from omni_schema.assemble_schema(%L, omni_vfs.local_fs(%L))', plan_db_conn, path)
```
yrashk, line 42: Would this approach mean we might not be able to ensure the order of migrations?

If a migration failed and proceeded further down migrations and only then went for sources, this would mean we'll come and try to re-apply the incremental migration.

Whereas if I got it right, the algorithm is to try migrations in order and when they error our, try to run sources.

extensions/omni_schema/src/make_plan.sql
```diff
@@ -0,0 +1,129 @@
...
+        if migration.migration_filename ~* 'migrations' then
+            -- migrations always execute only once
+            if not exists(select from omni_schema.migrations m where lower(m.name) = lower(omni_vfs.basename(migration.migration_filename))) 
+            then
+                return query select migration.migration_filename, migration.executed_statement::omni_sql.statement;
+            end if;
+        else
+            -- Non-migrations are picky, we error unless conditions are met
+            if not (
+                migration.executed_statement ~* 'select omni_python.install_requirements'
```
yrashk, line 74: is this particular line fragile? or good enough for now?

diogob: Yes, I was not happy with this. Was mostly a stop-gap solution. But we should review the approach to `omni_python` and other language loaders. Perhaps they should live in their own modules and we could invert the dependency, so they would use `omni_schema` to build a more refined plan.

extensions/omni_schema/src/make_plan.sql
diogob: It seems to me that the `assemble_schema` order will be preserved. But having it implicit is certainly not a good idea. Perhaps we could output an order from `assemble_schema` at least for now. 

The goal was to have something quick that would not require changes in `assemble_schema`. Later it might be better to just rewrite the logic inside `make_plan`.

extensions/omni_schema/docs/reference.md
diogob: :man_facepalming: some editing left-over

yrashk:
A note on replaceability:

```postgresql
omni_httpd=# create or replace procedure print() language plpgsql as $$ begin raise notice '%', current_setting('[test.me](https://test.me/)'); end; $$;
CREATE PROCEDURE
omni_httpd=# alter procedure print() set [test.me](https://test.me/) to 'hello';
ALTER PROCEDURE
omni_httpd=# call print();
NOTICE: hello
CALL
omni_httpd=# create or replace procedure print() language plpgsql as $$ begin raise notice '%', current_setting('[test.me](https://test.me/)'); end; $$;
CREATE PROCEDURE
omni_httpd=# call print();
NOTICE: 
CALL
```

This shows that things like altered settings vanish upon replacement. Which means that straight-out replacing objects is actually not lossless.

Reading documentation:

<blockquote>
When CREATE OR REPLACE PROCEDURE is used to replace an existing procedure, the ownership and permissions of the procedure do not change. All other procedure properties are assigned the values specified or implied in the command.
</blockquote>

The language of it is kind of vague but it may suggest this is intentional, or at least known.

PR #753, opened 2025-01-17
Problem: Executing migrations breaking transactions appart may lead to unexpected behaviour. (🚀 omni_sql 0.5.0)

diogob:
Since `omni_schema.assemble_schema` will execute sql files line by line and intersperse file contents based on dependency errors, this could change the semantics of transactions in files.

Solution: Keep transaction blocks intact when parsing statements. A parameter was introduced with a default value to keep backwards compatibility and it can be used to further analyse transactions line by line.


extensions/omni_sql/omni_sql.c
```diff
@@ -149,26 +149,58 @@ Datum raw_statements(PG_FUNCTION_ARGS) {
   rsinfo->setResult = tupstore;
 
   char *statement = PG_GETARG_CSTRING(0);
+  bool preserve_transactions = PG_GETARG_BOOL(1);
```
yrashk, line 152: `preserve_statement` should be checked for nullity.

And, as a matter of fact, so should be`statement` (my fault!)

extensions/omni_sql/tests/omni_sql/raw_statements.yml
```diff
@@ -26,3 +26,33 @@ tests:
...
+- name: with preserve transactions = true, a transaction and another statement
+  query: select *
+         from omni_sql.raw_statements(E'begin;Select True;select 1;end;select version()', true)
+  results:
+  - source: begin; Select True; select 1; end
+    line: 1
+    col: 1
+  - source: select version()
+    line: 1
+    col: 32
```
yrashk, line 58: A few ideas for extra tests:

*  what happens if we start a transaction but never end one?
* make sure `start transaction` syntax works
* same for `begin transaction ...`, `end transaction ...`

Also, what should we do with nested begin/end? They are not "valid" constructs, but they are valid syntactically.

extensions/omni_sql/omni_sql.c
```diff
@@ -149,26 +149,58 @@ Datum raw_statements(PG_FUNCTION_ARGS) {
...
+
+      if (preserve_transactions) {
+        // begin transaction
+        if (nodeTag(raw_stmt->stmt) == T_TransactionStmt && tx == NULL) {
+          tx = cmd;
+          tx_line = line;
+          tx_col = col;
+          // end transaction
+        } else if (nodeTag(raw_stmt->stmt) == T_TransactionStmt) {
+          Datum values[3] = {DirectFunctionCall2(textcat,
```
yrashk, line 180: Minor nitpick: commentary on what this concatenation does exactly may be useful for ease of reading.

extensions/omni_sql/omni_sql.c
```diff
@@ -149,26 +149,58 @@ Datum raw_statements(PG_FUNCTION_ARGS) {
...
-          Int32GetDatum(line), Int32GetDatum(col)};
-      bool isnull[3] = {false, false, false};
-      tuplestore_putvalues(tupstore, rsinfo->expectedDesc, values, isnull);
+      cmd = raw_stmt->stmt_len == 0
+                ? cstring_to_text(statement + raw_stmt->stmt_location + actual_start)
+                : cstring_to_text_with_len(statement + raw_stmt->stmt_location + actual_start,
+                                           raw_stmt->stmt_len - actual_start);
+
+      if (preserve_transactions) {
+        // begin transaction
```
yrashk, line 173: minor: maybe move `//begin` and `//end` inside of the `if/else` blocks to make it clearer to the reader in this case? Looking below it is not clear if we apply the scoping of the comments here consistently and is a bit confusing (at least to me)

extensions/omni_sql/tests/omni_sql/raw_statements.yml
diogob: :man_facepalming: - they should have this emoji on the reactions

extensions/omni_sql/omni_sql.c
diogob: `statement` check was there, I missed it too. Moved all these checks together to the top for the sake of readability

extensions/omni_sql/tests/omni_sql/raw_statements.yml
```diff
@@ -26,3 +26,80 @@ tests:
...
+
+- name: with preserve transactions = true and a rollbacked transaction
+  query: select *
+         from omni_sql.raw_statements(E'begin; Select True;\n select 1; rollback;', true)
+  results:
+  - source: begin; Select True; select 1; rollback
+    line: 1
+    col: 1
+
+- name: with preserve transactions = true and an unfinished transaction
```
yrashk, line 66: I am trying to figure out here if this is how it should work.

If I take this expression and it send it.. it's an unfinished transaction hanging there.

If we ungroup it, then the net effect is the same.

Should we actually error out here? And then, perhaps, error out if we encounter another begin in an open transaction?

yrashk: Totally special-casing here, but what if we had another optional parameter saying what to do with this situation: auto-rollback, leave as is, don't group?

diogob: I would avoid additional parameters until there is a concrete use case. The current implementation pushes the problem down the line. Raising an error would potentially save the user some headache.

diogob: I thought a bit more about this, and could not see any interesting use case for the unfinished transaction. So I decided to send a commit going down the simpler and safer approach and just error on the unfinished case.

diogob: I left this last change in a separate commit in case you rather drop it and keep the previous behavior.

PR #758, opened 2025-01-18
Problem: Fix retry logic and priority handling in assemble_schema

yadavshubham01:
Problem Overview:
In the current implementation of the assemble_schema function, once a failure occurs during the execution of a file, the process moves on to the next file, effectively ignoring the file's priority order. This leads to lower-priority files being executed before higher-priority ones, breaking the intended execution sequence. Additionally, when an error happens, revisiting failed files could lead to an infinite loop due to retrying without respect to previously failed files.

Solution:
To address the issue, we implemented the following changes:

1.  Retry Logic with Priority Preservation:

* We added a retry mechanism to the assemble_schema function, where files are retried up to a maximum of max_retries.
* Files are retried based on their priority, ensuring that higher-priority files are always executed before lower-priority ones, even 
  after a failure.
* We introduced a retry_count for each file to track how many times it has been retried, preventing an infinite loop in case of repeated failures.
2.  Priority-Based Execution Order:

* We ensured that the retry attempts respect the initial sorting order based on the priorities defined in omni_schema.languages and omni_schema.auxiliary_tools. The files are always executed in the correct order of priority, even after retries.
* The execution is handled sequentially within each file, and upon failure, the retries follow the original priority order rather than simply proceeding to the next file.
3. Error Handling and Logging:

* In case of failure during execution, we log the error and update the omni_schema_execution_status table with detailed error messages.
* If a file reaches its maximum retry limit, it is marked as permanently failed to avoid infinite loops.

### DIFF
--- a/extensions/omni_schema/src/assemble_schema.sql
+++ b/extensions/omni_schema/src/assemble_schema.sql
@@ -18,7 +18,6 @@ declare
     error_message                   text;
     error_detail text;
     try_again                       boolean = false;
-    try_again                       boolean = true;
     failed_files                    text[] = {};  
     retry_count                     int = 0;     
     max_retries                     int = 3;      
@@ -188,7 +187,7 @@ begin
 
                         -- If retry count is less than max retries, add file to retry queue
                         if retry_count < max_retries then
-                            failed_files := array_append(failed_files, rec.filepath);
+                            failed_files= array_append(failed_files, rec.filepath);
                         else
                             -- After max retries, mark as permanently failed
                             update omni_schema_execution_status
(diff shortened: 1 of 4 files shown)


yadavshubham01:
hey @yrashk please review this code like it need any other update then i will make change in tests of this

yrashk:
Original test cases fail with errors. Please ensure tests pass locally first.

yrashk:
Original test cases still fail with errors. Please ensure tests pass locally first, before submitting for a review.

yadavshubham01:
hey @yrashk get this error 
image

yrashk:
You need to install bison

yadavshubham01:
Hey @yrashk, I just ran the test case, but it still gives an error. Could you please review the logic I wrote and provide some feedback? It would really help!

PR #762, opened 2025-01-21
Problem: inja renderer uses C++ 20's ends_with

Vonng:
The ends_with is introduced by C++20, which is not available by default in distro like EL8

Solution: replace the function with a bespoke version to work around it


yrashk:
Thank you so much!

PR #779, opened 2025-02-07
Problem: omni_credential cookies are unexpirable

yrashk:
There is not a great amount of flexibility when it comes to cookies and therefore we can't limit their time, scope, etc.

Solution: make it possible to specify cookie parameters


yrashk:
🤦 I meant omni_session cookies

So much for writing messages at 5am :)

PR #783, opened 2025-02-10
Problem: omni_httpd request dispatch mechanism

yrashk:
It's flexible but requires replacing a dispatch function. That is not a great experience and is not very introspectible.

Solution: introduce URLPattern-based matching


yrashk:
Plan:

- [ ] priorities
- [ ] passing route tuple into the handler if expecting
- [ ] passing urlpattern named matches by parameter name? 🤔 or just let it use omni_web urlpattern facilities?
- [ ] (later?) speed up by pre-caching router relations but not views.

PR #785, opened 2025-02-10
The view omni_schema.role is not in 6NF 

diogob:
## Problem
The view omni_schema.role is not in 6NF making a diff between rows harder.

## Solution

Normalize it.
Note that there is a special case for the `role_setting` view.
Since each array item in `pg_db_role_setting.setconfig` is added or removed by a distinct `ALTER ROLE` command I created one line in `role_setting` for each config key.


yrashk:
Thank you! Looks good.

Just a thought I wanted to capture here: is this a problem for us that roles are cluster-wide? It means that meta will include this information in every database.

PR #786, opened 2025-02-11
Problem: omni_schema's assemble_schema execution order 

AP-51:
When a file execution fails, it previously used to proceed to the next available file. This led to higher priority failed files being executed after lower priority successful files. 

Solution: Add a retry counter column and keep track of how many retries a file has attempted. Also instead of continuing to the next file, instead let the loop naturally continue over the remaining files. Files are sorted using `auxillary.priority` and `languages.priority`.

### DIFF
--- a/extensions/omni_schema/src/assemble_schema.sql
+++ b/extensions/omni_schema/src/assemble_schema.sql
@@ -18,6 +18,8 @@ declare
     error_message                   text;
     error_detail text;
     try_again                       boolean = false;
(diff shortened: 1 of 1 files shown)


AP-51:
@yrashk could you give me some advice? I don't really understand how I'm supposed to run tests locally for this. My code is failing 4 tests, but I can't pinpoint where in the code and why.

yrashk:
https://github.com/omnigres/omnigres?tab=readme-ov-file#running-tests

PR #788, opened 2025-02-12
Problem: The view omni_schema.trigger is not in 6NF

diogob:
## Problem
The view omni_schema.trigger is not in 6NF making a diff between rows harder.

## Solution
Normalize it.

## TODO
- [x] trigger_enabled
- [ ] trigger_extension_dependency
- [x] tests

### DIFF
--- a/extensions/omni_schema/src/meta/catalog.sql
+++ b/extensions/omni_schema/src/meta/catalog.sql
@@ -1127,19 +1127,33 @@ create view trigger as
            t.id as relation_id,
            t_pgn.nspname::text as schema_name,
            pgc.relname::text as relation_name,
-           pg_trigger.tgname::text as name,
---            f.id as function_id,
-           case when (tgtype >> 1 & 1)::bool then 'before'
-                when (tgtype >> 6 & 1)::bool then 'before'
-                else 'after'
-           end as "when",
-           (tgtype >> 2 & 1)::bool as "insert",
-           (tgtype >> 3 & 1)::bool as "delete",
-           (tgtype >> 4 & 1)::bool as "update",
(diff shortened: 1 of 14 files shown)


extensions/omni_schema/src/meta/catalog.sql
```diff
@@ -1264,11 +1275,84 @@ create view trigger as
             on pgp.oid = tgfoid
 
     inner join pg_namespace f_pgn
-            on f_pgn.oid = pgp.pronamespace
+            on f_pgn.oid = pgp.pronamespace;
+
+create view trigger_when as
+    select trigger_id(t_pgn.nspname, pgc.relname, pg_trigger.tgname) as id,
+           case when (tgtype >> 1 & 1)::bool then 'before'
+                when (tgtype >> 6 & 1)::bool then 'before'
```
yrashk, line 1283: did you mean `instead` here?

yrashk:
Great! Some thoughts coverage-wise:

* Should we process `tgenabled`? (I cautiously  think yes?)
* Should we filter out `tgisinternal`? (I suspect so)
* Should we track `tgdeferrable` and `tginitdeferred`? (I think we should as this is often critical)
* Should we track `tgnargs` and `tgargs`? (I think not tracking is lossy)
* Should we track `tgattr`? (triggers can be column-specific)
* Should we track `tgqual`? (triggers can be only done when conditions are met, if we don't track it, we lose it)
* Should we track the `referencing` clause (`tgoldtable` and `tgnewtable`)? (I suspect we should because trigger would have an expectation for it)

yrashk:
Merging this for the base that we can work on.

PR #790, opened 2025-02-12
Problem: tracking schema evolution

yrashk:
Tracking schema evolution, especially in presence of data-native functions (fka _stored procedures_) is a non-trivial task:

* How do we prevent data and auxiliary information (such as presence of indices) loss?
* How do we ensure that schema is evolved as intended?
* How do we make sure the correct set of extensions at correct versions is used for every specific version?
* How do we sure data-native functions correspond the correct revision of the schema?
* and so on.

Solution: implement a robust tracking solution
following [RFC0001](https://github.com/omnigres/rfc/pull/1)


yrashk:
Also, we should probably exclude implicit constrain triggers (those that look like `RI_ConstraintTrigger..`) and capture constraints correctly. Naming is an implementation detail there.

yrashk:
What's still missing:

- [ ] excluding extension objects
- [ ] separating diffs by parents

PR #791, opened 2025-02-12
Problem: omni_schema meta does not include comments

yrashk:
This makes it lossy.

Solution: include comment view

It uses sum types to map comments to all kinds of types instead of creating a new view for every type. Time will show if this is a good choice.

It does not have full coverage yet.


yrashk:
Included in #766

PR #797, opened 2025-02-18
Problem: no response from false websocket_handler

sqwishy:
If `omni_httpd.websocket_handler()` returns false, no HTTP response is sent. And it seems like the worker just hangs for some reason -- it never responds to future requests -- even if the connection is closed by the peer.

However, if the handler returns null, it does send an HTTP response that does not upgrade the connection to a WebSocket.

But the comment for the handler indicates that any non-true value should be treated the same:

    If it returns anything but true, the connection is rejected.

Solution: send a response in any case

This change should send an HTTP even if false is returned and only upgrade if the handler returns true.

I'm still not sure why the worker seemed to not receive further requests when it failed to send a response here. (Though omni_httpd.stop and start would restart them.) Even when the connection was closed by the peer, I don't think the worker was closing its file descriptor for the socket. But this might have something to do with how the h2o library works that I'm not familiar with.

### DIFF
--- a/extensions/omni_httpd/tests/websocket.yml
+++ b/extensions/omni_httpd/tests/websocket.yml
@@ -1,0 +1,32 @@
+$schema: "https://raw.githubusercontent.com/omnigres/omnigres/master/pg_yregress/schema.json"
+instance:
+  config:
+    shared_preload_libraries: */env/OMNI_SO
+  init:
+  - set session omni_httpd.init_port = 0
+  - create extension omni_httpd cascade
+  - create extension omni_httpc cascade
+  # FIXME: waiting for two reloads is working around a startup bug in omni_httpd
+  - call omni_httpd.wait_for_configuration_reloads(2)
+
+tests:
+
+- name: websocket upgrade requests rejected by default
+  query: |
+    SELECT status
+         , (SELECT json_agg(json_build_object(h.name, h.value))
+              FROM unnest(response.headers) h
+             WHERE h.name != 'server') as headers
+      FROM omni_httpc.http_execute(
+             omni_httpc.http_request('http://127.0.0.1:' || (SELECT effective_port FROM omni_httpd.listeners) || '/',
+                                     'GET',
+                                     array[omni_http.http_header('connection', 'Upgrade'),
+                                           omni_http.http_header('upgrade', 'websocket'),
+                                           omni_http.http_header('sec-websocket-version', '13'),
+                                           omni_http.http_header('sec-websocket-key', 'dGhlIHNhbXBsZSBub25jZQ==')])) response
+  results:
+  # FIXME probably not the right status code
+  - status: 201
+    headers:
+    - connection: close
+    - content-length: 0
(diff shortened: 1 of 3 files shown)


sqwishy:
Hey man, hope you're doing well. I only _just now_ noticed the next/omni_httpd branch has a lot of recent updates and an open pull request. I didn't test if this was an issue in that branch yet, but it looks like the code for this part is the same. I'm not sure it would merge cleanly because the file has changed so much. So I don't know if you'd feel better merging this into that branch instead or just applying the change yourself because it's pretty trivial; assuming the rationale for this fix makes sense.

sqwishy:
Nevermind! I can't reproduce this on next/omni_httpd. 🙃

sqwishy:
Double nevermind! This does seem to happen in next/omni_httpd, I forgot to update the extension in postgres after recompiling.

yrashk:
Thank you for this contribution!

Do you think a pg_yregress test can be devised to ensure this behavior stays?

In other notes, `next/omni_httpd` would be a good target for this PR, along with a short note in the changelog in the `Fixed` section.

extensions/omni_httpd/tests/websocket.yml
```diff
@@ -0,0 +1,32 @@
...
+      FROM omni_httpc.http_execute(
+             omni_httpc.http_request('http://127.0.0.1:' || (SELECT effective_port FROM omni_httpd.listeners) || '/',
+                                     'GET',
+                                     array[omni_http.http_header('connection', 'Upgrade'),
+                                           omni_http.http_header('upgrade', 'websocket'),
+                                           omni_http.http_header('sec-websocket-version', '13'),
+                                           omni_http.http_header('sec-websocket-key', 'dGhlIHNhbXBsZSBub25jZQ==')])) response
+  results:
+  # FIXME probably not the right status code
+  - status: 201
```
yrashk, line 29: Perhaps 404?

sqwishy: So I think the normal behaviour is to ignore the upgrade header if the server does not want to upgrade the connection. That seems to be what some RFCs indicate and how nginx and caddy work and the [the MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) says:

> If it does not (or cannot) upgrade the connection, it ignores the Upgrade header and sends back a regular response (for example, a [200 OK](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200)).

So it might be reasonable where, if the websocket upgrade is declined by the database's handler function, request handler in the omni_httpd extension proceeds to handle the request as if it were a regular request and not an upgrade.

Another option is allowing the user defined handler function to provide a response and choose its own status code. This is a bit of scope creep but allows a handler to reply with 401 or something if it requires authentication and didn't receive any in the request.

But I guess technically a 201 response isn't incorrect. So maybe the test shouldn't check the status code at all.

yrashk: Very astute assessment. I guess we can start at the current behavior and expand later. 201 might be confusing, though as it is "Created". Maybe 204 No Data or 400s would be more appropriate for now

PR #810, opened 2025-02-23
Problem: omni_httpd leaks file descriptors

sqwishy:
WebSockets cause omni_httpd to leak file descriptors, mainly through the unix socket proxy tunnel thingy and but also not closing the TCP connection. Eventually, it runs out of file descriptors and cannot accept new connections. It can be particularly troublesome if accept() fails with "Too many open files" and gets stuck trying again and again failing each time indefinitely. I think better error handling and reporting would valuable in future work.

Solution: close file descriptors for opened sockets

1. `on_ws_message` might not call h2o_websocket_close on the TCP socket for a WebSocket

   `on_ws_message` would see `arg->opcode == WSLAY_CONNECTION_CLOSE` and `free(conn->data)` and assign it to NULL.

   Then `on_ws_message` would see `arg == NULL` but not enter the branch to call `h2o_websocket_close()` because data is NULL at that point.

   This only frees the data and closes the socket when when `arg == NULL`.

2. When a postgres connection process calls `websocket_send` from a SQL query, it needs to forward the message to whatever httpd worker holds the underlying connection for that WebSocket.

   In doing so it opens a socket but never closes it, it stores it in a collection to re-use it later but it does eventually leak and long running connection processes can run into problems for this reason.

   This change simply closes the socket each time and removes the collection where open sockets were stored. This does have negative performance impact in terms of speed but positively impacts performance as the process is not doomed to run out of file descriptors. :)

   The two `send()` calls have been replaced with one `sendmsg()` using scatter-gather but I'm not sure this helps with speed all that much. (There is a lot of noise when I try to measure this extension on my machine.)

   So there is certainly room for a more efficient design, maybe one that includes the collection but is length limited and closes least recent sockets.

3. Also in `on_ws_message`, the unix listen socket for accepting connections from `websocket_send` was not being closed. The unix socket is opened when the WebSocket is upgraded. With this change, the file descriptor is copied to a `on_ws_message_data_t` structure pointed to by the h2o connection where it can be closed by `on_ws_message` when the WebSocket is closed.

Finally, h2o provides uuids for connections through `const char *h2o_conn_get_uuid(h2o_conn_t *)`. It may be simpler to use that and remove the uuid generation and formatting code in this extension. As far as I know, once an HTTP connection is upgraded to a WebSocket, it cannot be used for any other HTTP requests. Even in HTTP2 which supports streaming, once it's upgraded, it stops being whatever it was and the connection is a WebSocket from then on. So from what I understand, it's not as if more than one WebSocket can run on one TCP connection; it can't be upgraded again in a separate stream. So if the WebSocket uuid is just the connection uuid, it would still remain unique across WebSockets.

### DIFF
--- a/extensions/omni_httpd/event_loop.c
+++ b/extensions/omni_httpd/event_loop.c
@@ -172,24 +172,30 @@ static inline void prepare_req_for_reprocess(h2o_req_t *req) {
   req->conn->ctx->proxy.client_ctx.max_buffer_size = H2O_SOCKET_INITIAL_INPUT_BUFFER_SIZE * 2;
 }
 
+typedef struct on_ws_message_data {
+  websocket_uuid_t ws_uuid;
+  int unix_socket;
+} on_ws_message_data_t;
+
 static void on_ws_message(h2o_websocket_conn_t *conn,
                           const struct wslay_event_on_msg_recv_arg *arg) {
-  char *uuid = (char *)conn->data;
-  if (conn->data != NULL && (arg == NULL || arg->opcode == WSLAY_CONNECTION_CLOSE)) {
+  on_ws_message_data_t *data = (on_ws_message_data_t  *)conn->data;
+
+  if (data != NULL && arg == NULL) {
     handler_message_t *msg = malloc(sizeof(*msg));
     msg->super = (h2o_multithread_message_t){{NULL}};
     msg->type = handler_message_websocket_close;
-    memcpy(msg->payload.websocket_close.uuid, uuid, sizeof(msg->payload.websocket_close.uuid));
+    memcpy(msg->payload.websocket_close.uuid, data->ws_uuid, sizeof(msg->payload.websocket_close.uuid));
 
     h2o_multithread_send_message(&handler_receiver, &msg->super);
 
-    // Remove the socket
-    struct sockaddr_un server;
-    websocket_unix_domain_socket(conn->data, &server, true);
+    // Close & remove the socket
+    close(data->unix_socket);
+    unlink_websocket_unix_socket(&data->ws_uuid);
+
     // Mark this connection as closed and dispose the UUID
-    if (arg == NULL) {
-      h2o_websocket_close(conn);
-    }
+    h2o_websocket_close(conn);
+
     free(conn->data);
     conn->data = NULL;
     return;
@@ -203,7 +209,7 @@ static void on_ws_message(h2o_websocket_conn_t *conn,
     msg->payload.websocket_message.message = (uint8_t *)malloc(arg->msg_length);
     msg->payload.websocket_message.message_len = arg->msg_length;
     memcpy(msg->payload.websocket_message.message, arg->msg, arg->msg_length);
-    memcpy(msg->payload.websocket_message.uuid, uuid, sizeof(msg->payload.websocket_open.uuid));
+    memcpy(msg->payload.websocket_message.uuid, data->ws_uuid, sizeof(msg->payload.websocket_open.uuid));
 
     h2o_multithread_send_message(&handler_receiver, &msg->super);
   }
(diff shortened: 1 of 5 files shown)


sqwishy:
I think there are four(?) places in `event_loop.c:on_message` where I wasn't super sure about how to handle errors there. So I left some funny comments in the commit about them as a note to get feedback on what would be appropriate.

If there isn't a clear way to log or handle errors there, I can just remove the comments. But it would be nice to replace the comments with even some logging or something. I'm just not super sure what the options are there since it's happening in the http thread.

yrashk:
On the source code formatting: https://github.com/omnigres/omnigres/blob/master/HACKING.md#source-code-formatting

yrashk:
Thank you!

I like the idea of re-using `h2o_conn_get_uuid` from H2O. While it probably doesn't change much for us, it just reduces the amount of code we must maintain by a little bit.

extensions/omni_httpd/event_loop.c
```diff
@@ -324,19 +337,35 @@ static void on_message(h2o_multithread_receiver_t *receiver, h2o_linklist_t *mes
       break;
     }
     case MSG_KIND_UPGRADE_TO_WEBSOCKET: {
-      h2o_websocket_conn_t *websocket_conn =
-          h2o_upgrade_to_websocket(req, reqmsg->ws_client_key, reqmsg, on_ws_message);
-      websocket_conn->data = malloc(sizeof(reqmsg->ws_uuid));
-      memcpy(websocket_conn->data, reqmsg->ws_uuid, sizeof(reqmsg->ws_uuid));
+      on_ws_message_data_t *data = malloc(sizeof(on_ws_message_data_t));
+      // if (!data) ???
```
yrashk, line 341: Allocation failure? Well, we can't proceed  from there so I guess we just log a WARNING and abort?

extensions/omni_httpd/event_loop.c
```diff
@@ -324,19 +337,35 @@ static void on_message(h2o_multithread_receiver_t *receiver, h2o_linklist_t *mes
...
+      if (   websocket_unix_socket_path(&server_addr, &reqmsg->ws_uuid) < 0
+          || (data->unix_socket = socket(AF_UNIX, SOCK_STREAM, 0)) < 0) {
+        /* todo log errors? */
+        free(data);
+        goto done;
+      }
+
+      if (   bind(data->unix_socket, (struct sockaddr *)&server_addr, sizeof(struct sockaddr_un)) < 0
+          || listen(data->unix_socket, SOMAXCONN) < 0) {
+        /* todo log errors? */
```
yrashk, line 357: just as above, log a WARNING?

extensions/omni_httpd/omni_httpd.c
```diff
@@ -559,44 +559,59 @@ static inline bool sh_uuid_eq(struct sockethash_hash *tb, const pg_uuid_t a, con
 Datum websocket_send(PG_FUNCTION_ARGS, int8 kind) {
   if (PG_ARGISNULL(0)) {
     ereport(ERROR, errmsg("socket can't be null"));
+    PG_RETURN_VOID();
```
yrashk, line 562: No need for this as `ereport(ERROR)` is non-terminating.

extensions/omni_httpd/omni_httpd.c
```diff
@@ -559,44 +559,59 @@ static inline bool sh_uuid_eq(struct sockethash_hash *tb, const pg_uuid_t a, con
...
   }
   if (PG_ARGISNULL(1)) {
     ereport(ERROR, errmsg("data can't be null"));
+    PG_RETURN_VOID();
```
yrashk, line 566: Same as above.

extensions/omni_httpd/omni_httpd.c
```diff
@@ -559,44 +559,59 @@ static inline bool sh_uuid_eq(struct sockethash_hash *tb, const pg_uuid_t a, con
...
   }
 
-  static sockethash_hash *sockethash = NULL;
```
yrashk, line 567: If we are removing this, should we remove the type definition as well?

yrashk (approved):
Were you able to measure the performance impact of closing the sockets?

sqwishy:
> Were you able to measure the performance impact of closing the sockets?

There was a lot of run to run variance but the test I ran took 14 seconds before and 18 now in the worst case. I think that was testing like 500k send and receives to an echo server. So it's not great, but it's not so bad that it is unusable; which I did want to be sure of.

extensions/omni_httpd/event_loop.c
sqwishy: I wasn't sure how to log it. Just print to stderr?

Everything else logs to postgres by this call is running in the h2o thread and does not have access to anything postgres directly and probably shouldn't as I haven't seen anything to suggest that it is threadsafe?

sqwishy:
Logging and formatting should be correct now. Thanks for your feedback. Let me know if there's anything else.

PR #822, opened 2025-03-09
fix: omni_vfs.file_info for table_fs on root

### LINKED ISSUE #796: Problem: omni_vfs.file_info for table_fs on root
`omni_vfs.file_info(table_fs, '/')` would return a `null`. This is different from what `local_fs` would do.  

This is an inconsistent and surprising behavior.

Proposed solution: make sure it returns a non-null value describing it as a directory

Ayushjhawar8:
/claim #796
close #796

### DIFF
--- a/extensions/omni_vfs/src/file_info_table_fs.sql
+++ b/extensions/omni_vfs/src/file_info_table_fs.sql
@@ -3,9 +3,17 @@ create function file_info(fs table_fs, path text) returns omni_vfs_types_v1.file
     language sql set search_path to omni_vfs
 as
 $$
-
+-- Special case for root directory
 with
-    match(id) as (select table_fs_file_id(fs, path))
+    match(id) as (
+        select 
+            case 
+                when omni_vfs.canonicalize_path(path, absolute => true) = '/' then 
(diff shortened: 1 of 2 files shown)


yrashk:
Thank you for the submission!

Please follow the [contribution guidelines](https://github.com/omnigres/omnigres/blob/master/CONTRIBUTING.md) and write tests as well.

extensions/omni_vfs/src/file_info_table_fs.sql
```diff
@@ -15,7 +23,7 @@ select
 from
     table_fs_files                f
     inner join match              m on f.id = m.id
-    inner join table_fs_file_data d on m.id = d.file_id
+    left join table_fs_file_data  d on m.id = d.file_id
```
yrashk, line 26: Do I get this correctly that you're changing this to a left join to accommodate for the fact that the root entry does not / might not exist? If so, do you think it'd be worthwhile to explain this briefly in a comment?

yrashk:
CleanShot 2025-03-09 at 08 27 57@2x

As per acceptance criteria and guidelines, please ensure tests work locally before submitting a non-Draft PR.

PR #825, opened 2025-03-10
Fix file_info to return valid metadata for root directory in table_fs

### LINKED ISSUE #796: Problem: omni_vfs.file_info for table_fs on root

deekshatomer:
/claim #796 
/fixes #796 

This PR fixes issue [#796](https://github.com/omnigres/omnigres/issues/796), ensuring that omni_vfs.file_info(table_fs, '/') returns proper metadata instead of NULL.

Problem:

Previously, file_info(table_fs, '/') returned NULL, unlike local_fs, which provided proper metadata.
This inconsistency led to unexpected behavior when querying the root directory of table_fs.
Solution:

Updated the file_info function to ensure / in table_fs always returns a valid directory entry.
Ensured size, created_at, accessed_at, and modified_at return meaningful values.
Adjusted query logic to prevent NULL results for root.
Testing Done:

Ran SELECT * FROM omni_vfs.file_info(...) and verified correct timestamps and metadata for /.
Ensured no regressions in behavior for other paths.
Screenshot of Fix


yrashk:
Please follow the guidelines and acceptance criteria as per bounty description in #796: tests in omni_extensions/omni_vfs/tests/, follow the contribution workflow.

Also, unnecessary syntax changes in the code (uppercasing).

Issue #771, opened 2025-01-31
Problem: omni_credentials file store is limited to local file system


yrashk:
This pierces through our model of having VFS in place. I think the reason why this happened is because at the time, `omni_vfs` didn't have file writing capabilities and `omni_credentials` model depended on that functionality to enable this workflow.

Proposed solution: migrate to omni_vfs

AgCaliva:
Humm sory if im wrong but why there is always same people with multiple accounts making PR that fail on building stages? They are bots? If thats the case then this is still available? Thats why it is still open isnt it?

oliverqx:
@AgCaliva as of now, build stage fails because omni_credentials branch needs merging with main. Have not looked into luffy's pr long enough to know why its not building. I'd say its still open because the current possible solution that meets all requirements feels very flaky. I am currently diving into omnigres as a whole to provide a much more substantial resolutiion

AgCaliva:
@oliverqx yes you are real, but luffy and other of his alts he uses seems very like AI bots, i see you are almost there at this one, but in the other issues luffy and others also spammed crap suspicious PRs

yrashk:
The bounty program has been permanently closed.

PR #833, opened 2025-03-19
Problem: omni_credentials file store is limited to local file system

oliverqx:
/claim #771

Solution: migrate to omni_vfs as recommended #771


extensions/omni_credentials/CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.3.0] - 2025-3-17
```
yrashk, line 8: Please follow the format as described and insert changes into 0.2.0 as that's the target release (you can see it is TBD)

extensions/omni_credentials/src/instantiate_file_store.sql
```diff
@@ -44,7 +78,38 @@ begin
...
+        into json_content
+        from encrypted_credentials;
+
+        perform omni_vfs.write(
+            omni_var.get_session('fs', null::omni_vfs.remote_fs),
+            filename,
+            convert_to(json_content::text, 'utf8')
+        );
+
+        return TRUE;
```
yrashk, line 108: please use lowercase

extensions/omni_credentials/src/instantiate_file_store.sql
```diff
@@ -18,17 +18,51 @@ begin
...
+        file_contents text;
     begin
         if filename not like '/%' then
             filename := current_setting('data_directory') || '/' || filename;
         end if;
-        perform pg_stat_file(filename);
+        
+        select convert_from(
+            omni_vfs.read(
+                omni_var.get_session(
```
yrashk, line 30: I am not sure why you need a session here, you can pass `fs` into it.

extensions/omni_credentials/src/instantiate_file_store.sql
```diff
@@ -44,7 +78,38 @@ begin
...
+                'kind', kind,
+                'principal', principal,
+                'scope', scope
+            )
+        )
+        into json_content
+        from encrypted_credentials;
+
+        perform omni_vfs.write(
+            omni_var.get_session('fs', null::omni_vfs.remote_fs),
```
yrashk, line 103: same on the session usage as above

extensions/omni_credentials/src/instantiate_file_store.sql
```diff
@@ -44,7 +78,38 @@ begin
     insert into credential_file_stores (filename) values (instantiate_file_store.filename);
 
     perform credential_file_store_reload(filename);
-    execute format('copy encrypted_credentials to %L', filename);
+
+    create or replace function update_file(filename text) returns boolean
```
yrashk, line 82: Maybe we can find a slightly better name for this function? What do you think about `update_credentials_file`?

extensions/omni_credentials/tests/credentials_file_store.yaml
```diff
@@ -1,49 +1,107 @@
...
+  - create extension omni_vfs cascade
+  - create extension omni_var cascade
   - create extension omni_credentials cascade
   - create schema creds
   # FIXME: we use PATH because we can't set env variables yet
   - select omni_credentials.instantiate('creds', env_var => 'PATH')
   - create schema creds1
   # FIXME: we use PATH because we can't set env variables yet
   - select omni_credentials.instantiate('creds1', env_var => 'PATH')
+  - select omni_var.set_session(
```
yrashk, line 13: see above on the use of session

extensions/omni_credentials/src/instantiate_file_store.sql
```diff
@@ -18,17 +18,51 @@ begin
...
-        execute format('copy __new_encrypted_credentials__ from %L', filename);
-        raise notice '%', pg_read_file(filename);
+
+        insert into __new_encrypted_credentials__ (name, value, kind, principal, scope)
+        select
+            cred->>'name' AS name,
+            decode(cred->>'value', 'base64') AS value,
+            coalesce(
+                nullif(cred->>'kind', '')::credential_kind,
+                'credential'::credential_kind
```
yrashk, line 54: These defaults duplicate the defaults on the credentials table, don't they? Would it make sense for the table to enforce these defaults?

oliverqx: ah yes, you're right. Slipped my mind that postgres itself would have them as defaults

oliverqx:
@yrashk just addressed the changes except for the fs session variable one. I did try passing fs through, however an issue arises when creating the `file_store_credentials_update` trigger as it cannot take in any parameters in a clean way. I kinda wanted to avoid using TG_ARGV. Is there any other way you'd recommend doing it? should i just use TG_ARGV?

yrashk:
@oliverqx good point on the trigger; let me think about it. it's just that session vars won't really work in that case anyway as they'd have to be established on every new session – for every store.

oliverqx:
@yrashk i think i was able to solve it. I did the following:

1. expand `credential_file_stores` table to hold filename and vfs metadata (type, mount, connstr, constructor.
2. create `register_file_store(fs, filename)` function which will add a row to the `credential_file_stores`table, saving filename to and its associated vfs metadata.
3. update functions to retrieve vfs from `credential_file_stores` table using only filename

This makes it possible to dynamically create the right vfs, while allowing each credential file store to have a different vfs.

oliverqx:
@yrashk i think i was able to solve it by doing the following:

expand credential_file_stores table to hold filename and vfs metadata (type, mount, connstr, constructor, id)
create register_file_store(fs, filename) function which will add a row to the credential_file_storestable, saving filename and its associated vfs metadata.
update functions to retrieve vfs from credential_file_stores table using only filename
With this solution, its possible to:

file_store_credentials_update trigger works as expected because table data can be accessed at runtime/execution.
dynamically create the expected vfs, allowing each credential file store to have a different vfs.
and

UX is also better, as users don't need to pass in the vfs after registration, only the credential file store's name

oliverqx:
@yrashk i think i was able to solve it by doing the following:

1. expand `credential_file_stores` table to hold filename and vfs metadata (type, mount, connstr, constructor, id)
2. create `register_file_store(fs, filename)` function which will add a row to the `credential_file_stores` table, saving filename and its associated vfs metadata.
3. update functions to retrieve vfs from `credential_file_stores` table using only filename

With this solution:

1. , its possible to dynamically create the expected vfs, allowing each credential file store to have a different vfs.
2.`file_store_credentials_update` trigger works as expected because table data can be accessed at runtime/execution.
3. UX is also better, as users don't need to pass in the vfs after registration, only the credential file store's name

oliverqx:
I created a test suite for each vfs type. Its pretty repetitive but I don't know how to dry since each test suite follows the same structure but applies to different VFS types, and declarative code doesn't allow easy abstraction like loops or functions. I did try looking for loops or functions in other test suites, but couldn't find any

oliverqx:
@yrashk just wanted to ping you to see if you have any thoughts, would appreciate input. 

I was thinking I could save VFS under a JSON object so theres only one extra column to be added to `credential_file_stores`

yrashk:
Thank you for the ping. The issue with omni_vfs mapping is that their details are different – some have mounts, some have IDs. One one to join is them is to have a [sum type](https://docs.omnigres.org/omni_types/sum_types/). This introduces a limitation to dynamically added VFSs though and this is why it hasn't been done yet – trying to comprehend the impact.

yrashk:
I have an idea. What if during file store instantiation, we created a function, say `credential_vfs()` that returned the `pg_typeof(fs)` passed into it (something like `execute format($$create function credential_vfs() returns %I return %s;$, pg_typeof(fs),fs::text)` and then we called it from the trigger?

oliverqx:
I think doing it that way introduces a limits the amount of vfs's you can instantiate to one, because as soon as you instantiate another one the function is overwritten. Say, if you instantiate a table and then a local one, when the trigger runs, the proposed credential_vfs function would return pg_typeof(fs) of the latest one (local).

However that did push me into this idea:

What if have each vfs define two more functions:
1. `fs_info() returns json`
        this function would return all the info needed to call omni_vfs.{type}_fs('something') the given type of fs. 
        so instantiation_info(omni_vfs.table_fs('test_table_fs') would return { filesystem_name: 'test_table_fs' }
2. `{vfs_type}.initiate_fs(info jsonb) returns vfs_type`
        this function would take care of parsing the given jsonb, and if valid, would return the correct `vfs_type`

oliverqx:
I think doing it that way introduces a limits the amount of vfs's you can instantiate to one, because as soon as you instantiate another one the function is overwritten. Say, if you instantiate a table and then a local one, when the trigger runs, the proposed credential_vfs function would return pg_typeof(fs) of the latest one (local).

However that did push me into this idea:

What if have each vfs define two more functions:
1. `fs_info(fs) returns json`
        this function would return all the info needed to call the vfs initiation function. 
        so fs_info(omni_vfs.table_fs('test_table_fs')) would return { filesystem_name: 'test_table_fs' }
2. `omni_vfs.{vfs_type}_dynamic_write_to_file(info jsonb, filename text, content bytea) returns bool`
        this function would take care of parsing the given jsonb, and if valid, then would call `omni_vfs.write`

with that setup, when instantiating a credential filestore, the credential_file_stores table would store:
1. fs_type (text) which comes from pg_typeof(fs)
2. fs_info (jsonb) which comes from omni_vfs.table_fs_info(fs)
3. filename (text)

then the update_credentials_file function could look like this:

```
    create or replace function update_credentials_file(filename text) returns bigint
        language plpgsql
    as
    $code$
    declare
        json_content jsonb;
        fs_type text;
        fs_info jsonb
    begin
// 1. get info from credential_file_stores
        select 
            credential_file_stores.fs_type, 
            credential_file_stores.fs_info,
        into fs_type, fs_info
        from credential_file_stores 
        where credential_file_stores.filename = update_credentials_file.filename;

// 2. prepare content to store in file
        select jsonb_agg(
            jsonb_build_object(
                'name', name,
                'value', replace(encode(value, 'base64'), E'\n', ''),
                'kind', kind,
                'principal', principal,
                'scope', scope
            )
        )
        into json_content
        from encrypted_credentials;

// 3. write to file
  execute format('perform omni_vfs.%I_dynamic_write_to_file(
                fs_info
                filename,
                convert_to(json_content::text, 'utf8')
            );
  ', fs_type)
```

yrashk:
Thank you for the creative thinking! You are right that if you instantiate the credential store the same way, it'll overwrite (unless it will fail to instantiate sooner) – but I am a bit hesitant to change VFS API just yet. We already use a lot of dynamically generated code, and doing more of it is testing it.

Here are two branches of thinking that I have:

**Short/medium term**: what if when we instantiate a store, the function name will be unique, and the trigger will use that unique name (since we generate trigger code anyway)? 

**Long-term**: I am exploring the feasibility and practicality of flipping the way we access virtual file systems. Currently, they are built as APIs to _sort of_ mimic existing file access APIs. However, if we embrace "database as the interface" principle, then all file systems should be available through (updateable) views. This is definitely a large topic worth researching – what are the benefits and what are the downsides (fixable and unfixable). I don't think it should directly impact this particular PR (it's way out of of the scope), but I think it's worth keeping in mind. I've posted a note about this on the `#omni_ vfs` channel on Discord.

oliverqx:
@yrashk i attempted your short term solution, and its working!

I actually theorized going down this route but couldn't wrap my head around how the vfs would be 'initiated' in a clean way, till I realized its just a composite type and that's exactly what omni_vfs.vfs(x) returns. I kept thinking I needed to get back x to then call omni_vfs.vfs(x) if that makes sense. Have been studying postgres lately though. Its so much more flexible than I imagined

To contribute something relevant to the Long-Term vfs design conversation I definitely need more experience and a better understanding of omnigres as a whole, so it might take a bit

Anyway, would appreciate some feedback, all tests are running alright. I do think this was the cleanest way to go

yrashk:
Thank you. I will review this shortly! Failures on master can be avoided by ensuring a full rebase off master. I might need to do that on next/omni_credentials.

Quick thought that I had was whether it would make sense to insert the value right into generated values instead of keeping a run-time function that calculates the name.

oliverqx:
@yrashk I added a column to credential_file_stores called function_name, however, wdyt about just naming the function with  filename?

yrashk:
> wdyt about just naming the function with filename?
can you elaborate on this?

oliverqx:
Backtracking that comment sorry. My original idea was to have the function name be named after the given filename, this way an additional column to credential_file_stores is not necessary. However, the filename needs to be sanitized, so prepending with `credential_vfs` or not make much of a difference

yrashk:
Function names are limited to 63 characters – `name` type.

```
=# \dT name
                       List of data types
   Schema   | Name |                 Description                 
------------+------+---------------------------------------------
 pg_catalog | name | 63-byte type for storing system identifiers
```

oliverqx:
I left it that way because postgres itself should automatically truncate it, but I just pushed a change that does it manually and a test case

oliverqx:
@yrashk friendly ping about my pr as ive seen another attempt open. I think my attempt already enables vfs for credentials while supporting dynamically added vfs. Function name size is now handled properly

yrashk:
Apologies for the delay, was travelling. Looking into it.

yrashk:
I think the only visible point of hesitation I have at this moment is the introduction of function into the mix. Do you think it's worth considering just doing what that function expands to in places where we read out the name of the function and execute it? 

My perception is that currently we just have an extra level of indirection in code generation – first we `format()` the function we need and then we `format()` the call to it. Why not short-circuit it and do `format()` once when we need it? WDYT?

yrashk:
Perhaps the flakiness in 57686c4 would disappear if we use the approach suggested in https://github.com/omnigres/omnigres/pull/833#issuecomment-2910817916?

oliverqx:
I think that would limit amount of vfs to one again because when a new filestore is initiated the fs passed to instantiate_file_store is replaced. Creating the function with a unique name makes it permanent and also saves the value of fs at creation time. So if a new fs is passed, the trigger can still access the old fs because its stored in the function definition

oliverqx:
@yrashk I do like this version much more, it even offers a better dx. Function creation through formatting added some friction for new devs reading the code. 

also just merged!

yrashk:
Thank you so much for working on this! ❤️

oliverqx:
thanks!

PR #834, opened 2025-03-19
Problem: omni_vfs.file_info for table_fs on root

oliverqx:
/claim #796 

Problem takes place because the file_info function was using an inner join with table_fs_file_data.

Since directories don't have entries in this table, the query returned no results (all null values) when attempting to retrieve directory information.

Solution: Changed the query to use left join instead of inner join when joining to table_fs_file_data. This ensures that directories are included in the results even without corresponding data entries on table_fs_file_data, preserving the kind field while allowing data-specific fields to be null where appropriate.

This change makes it possible to call `omni_vfs.file_info(fs, directory)` on any directory, not just root

### DIFF
--- a/extensions/omni_vfs/tests/table_fs.yml
+++ b/extensions/omni_vfs/tests/table_fs.yml
@@ -428,3 +428,9 @@ tests:
     results:
     - data1: 0x75706461746564
       data2: 0x7365636f6e64
+
+- name: can get root directory info
+  query: select size > 0 as non_zero, kind from omni_vfs.file_info(omni_vfs.table_fs('fs'), '/')
+  results:
+  - non_zero: true
+    kind: dir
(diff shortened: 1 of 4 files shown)


extensions/omni_vfs/tests/table_fs.yml
```diff
@@ -226,6 +226,22 @@ tests:
...
+    steps:
+    - name: create directory
+      query: |
+        select omni_vfs.write(omni_vfs.table_fs('fs'), '/to/test', 'hello world', create_file => true)
+    
+    - name: get directory info
+      query: select kind from omni_vfs.file_info(omni_vfs.table_fs('fs'), '/to')
+      results:
+      - kind: dir
+
```
yrashk, line 244: Should we have a test on a non-existing dir?

oliverqx: its being created indirectly right above, during the first step of that test. I thought it would be nice to make sure it works not only for root but for any directory

yrashk: I meant to suggest an additional test case to ensure that we don’t return something incorrect if the directory doesn’t actually exist

oliverqx: ah okay makes total sense yes

Issue #823, opened 2025-03-10
Problem: omni_schema.dependency and omni_schema.acl views are slow


yrashk:
**We're closing down our bounty program; this issue is no longer open for new submissions.**

This gravely impacts some of our functionality like schema diffing as we scan through all these views and `dependency` and `acl` views are the worst offenders.

Proposed solution: find a way to significantly optimize the queries

One thing is to see what else can be done for the queries themselves – at the very least in `dependency` and `acl` and maybe even other meta views. 

https://github.com/omnigres/omnigres/blob/master/extensions/omni_schema/src/meta/catalog.sql#L76

yrashk:
/bounty $1000

Additional acceptance criteria: 

* need at least 2x **measured** (collect the before and after times) performance improvement for the worst offending views. The submission with the best improvement ratio wins. In the ideal world, we want to push for 5x and more. `dependency` often takes 250-400ms+ and that's not great.
* if the solution is non-obvious from reading the final code, please document it

harsh-791:
@yrashk can i work on this issue?

webbdays:
Hi,
@yrashk
what is the status of existing pr's?
Are they not solving issue?
Kindly let us know.
Thank you very much.

yrashk:
@rishi-jat – We're closing down our bounty program; this issue is no longer open for new submissions.

We're reviewing contending solutions at the moment.

ntoledo319:
/attempt #823

Plan:
- Reproduce baseline timings (dependency + acl) with EXPLAIN (ANALYZE, BUFFERS) and repeated measurements.
- Identify primary cost centers (row explosion, join predicates, repeated evaluation, materialization).
- Implement optimizations with correctness preserved; add regression test for any correctness bug fixed.
- Produce bench artifacts + documentation; run full local tests; bump versions.txt + CHANGELOG; open PR with /claim #823.

yrashk:
@ntoledo319 please note that as per https://github.com/omnigres/omnigres/issues/823#issuecomment-3638334349 we are no longer taking submissions.

PR #836, opened 2025-03-22
dynamically refresh dependency and acl. #823

benqethan:
/claim #823

### DIFF
--- a/extensions/omni_schema/src/meta/catalog.sql
+++ b/extensions/omni_schema/src/meta/catalog.sql
@@ -2093,298 +2093,429 @@ create view obj_object_id as
         pg_language l;
 --- TODO: handle the rest of the cases (^^)
 
-create view dependency as
-    with
-        pre as materialized (
-            -- relation
-            select
-                relation_id(ns.nspname, c.relname)::object_id as id,
-                d                                             as dependency
-            from
-                pg_depend               d
-                inner join pg_class     c
-                           on c.oid = d.objid and d.classid = 'pg_class'::regclass and c.relkind != 't'
-                inner join pg_namespace ns on ns.oid = c.relnamespace
-            where
-                d.objsubid = 0
-            union all
-            -- callable
-            select
-                function_id(ns.nspname, p.proname, _get_function_type_sig_array(p))::object_id as id,
-                d                                                                              as dependency
-            from
-                pg_depend               d
-                inner join pg_proc      p on p.oid = d.objid and d.classid = 'pg_proc'::regclass
-                inner join pg_namespace ns on ns.oid = p.pronamespace
-            where
-                d.objsubid = 0
-            union all
-            -- column
-            select
-                column_id(ns.nspname, c.relname, a.attname)::object_id as id,
-                d                                                      as dependency
-            from
-                pg_depend               d
-                inner join pg_class     c on c.oid = d.objid and d.classid = 'pg_class'::regclass
-                inner join pg_attribute a on a.attrelid = c.oid and a.attnum > 0
-                inner join pg_namespace ns on ns.oid = c.relnamespace
-            where
-                d.objsubid != 0
-            union all
-            --- columns also depend on what tables depend on
-            select
-                column_id(ns.nspname, c.relname, a.attname)::object_id as id,
-                (('pg_attribute'::regclass, c.oid, a.attnum, d.refclassid, d.refobjid, d.refobjsubid,
-                  d.deptype)::pg_depend)                               as dependency
-            from
-                pg_attribute            a
-                inner join pg_class     c on c.oid = a.attrelid and c.reltype != 0
-                inner join pg_namespace ns on ns.oid = c.relnamespace
-                inner join pg_depend    d on d.objid = c.oid and d.classid = 'pg_class'::regclass
-            where
-                a.attnum > 0
-            union all
-            -- cast
-            select
-                cast_id(st_ns.nspname, st.typname, tt_ns.nspname, tt.typname)::object_id as id,
-                d                                                                        as dependency
-            from
-                pg_depend               d
-                inner join pg_cast      c on c.oid = d.objid and d.classid = 'pg_cast'::regclass
-                inner join pg_type      st on st.oid = c.castsource
-                inner join pg_type      tt on tt.oid = c.casttarget
-                inner join pg_namespace st_ns on st_ns.oid = st.typnamespace
-                inner join pg_namespace tt_ns on tt_ns.oid = tt.typnamespace
-            where
-                d.objsubid = 0
-            -- type
-            union all
-            select
-                type_id(ns.nspname, resolved_type_name(t))::object_id as id,
-                d                                                     as dependency
-            from
-                pg_depend               d
-                inner join pg_type      t on t.oid = d.objid and d.classid = 'pg_type'::regclass
-                inner join pg_namespace ns on ns.oid = t.typnamespace
-            ---- not all types were properly connected before Postgres 17 (arrays)
-            union all
-            select
-                type_id(ns.nspname, resolved_type_name(t))::object_id as id,
-                d                                                     as dependency
-            from
-                pg_type                 t
-                inner join pg_namespace ns on ns.oid = t.typnamespace
-                inner join pg_depend    d on d.objid = t.typelem and d.classid = 'pg_type'::regclass
-            where
-                (current_setting('server_version_num')::int / 10000) < 17 and
-                t.typelem != 0
-            ---- not all types were properly connected before Postgres 17 (relations)
-            union all
-            select
-                type_id(ns.nspname, resolved_type_name(t))::object_id as id,
-                d                                                     as dependency
-            from
-                pg_type                 t
-                inner join pg_namespace ns on ns.oid = t.typnamespace
-                inner join pg_depend    d on d.objid = t.typrelid and d.classid = 'pg_class'::regclass
-            where
-                (current_setting('server_version_num')::int / 10000) < 17 and
-                t.typrelid != 0
-            ---- not all types were properly connected before Postgres 17 (relation arrays)
-            union all
-            select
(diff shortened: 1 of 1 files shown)


benqethan:
2 MATERIALIZED VIEWs are created to serve the dependency and acl view. The  MATERIALIZED VIEWs will be refreshed only there is change detected. 

- Below is the test result for dependency: (dependency_co is the new one). Please notes that the new view removed duplicate dependency and implicit dependency. 
```
omnigres=# \echo old_dep;
old_dep;
omnigres=# \echo :old_dep
select count(*) from omni_schema.dependency
    where dependent_on operator(omni_schema.=) omni_schema.extension_id('omni_schema')::omni_schema.object_id
    and omni_types.variant(id) = 'omni_schema.operator_id'::regtype;
omnigres=# \echo :co_dep
select count(*) from omni_schema.dependency_co 
    where dependent_on operator(omni_schema.=) omni_schema.extension_id('omni_schema')::omni_schema.object_id
    and omni_types.variant(id) = 'omni_schema.operator_id'::regtype;
omnigres=# :old_dep
 count 
-------
 83544
(1 row)

Time: 4962.267 ms (00:04.962)
omnigres=# :old_dep
 count 
-------
 83544
(1 row)

Time: 4925.130 ms (00:04.925)
omnigres=# :co_dep
NOTICE:  Checking materialized view: dependency_pre_mv, last_max_xid: 3501, current_max_xmin: 3501
NOTICE:  Materialized view dependency_pre_mv is not stale. No refresh needed.
 count 
-------
     6
(1 row)

Time: 95.118 ms
omnigres=# CREATE TABLE xxx (
    test_col INTEGER
);
CREATE TABLE
Time: 49.862 ms
omnigres=# :co_dep
NOTICE:  Checking materialized view: dependency_pre_mv, last_max_xid: 3501, current_max_xmin: 3508
NOTICE:  Materialized view dependency_pre_mv is stale. Refreshing...
NOTICE:  Materialized view dependency_pre_mv refreshed successfully.
 count 
-------
     6
(1 row)

Time: 1174.883 ms (00:01.175)
omnigres=# :co_dep
NOTICE:  Checking materialized view: dependency_pre_mv, last_max_xid: 3508, current_max_xmin: 3508
NOTICE:  Materialized view dependency_pre_mv is not stale. No refresh needed.
 count 
-------
     6
(1 row)

Time: 96.618 ms
omnigres=# :co_dep
NOTICE:  Checking materialized view: dependency_pre_mv, last_max_xid: 3508, current_max_xmin: 3508
NOTICE:  Materialized view dependency_pre_mv is not stale. No refresh needed.
 count 
-------
     6
(1 row)

Time: 91.342 ms
```

- Below is the test result for acl: (acl_co is the new one)
```
omnigres=# select count(*) from acl;
 count 
-------
 14298
(1 row)

omnigres=# REVOKE DELETE ON xxx FROM omni_test;
REVOKE
Time: 18.032 ms

omnigres=# select count(*) from acl_co;
NOTICE:  Checking materialized view: acl_mv, last_max_xid: 0, current_max_xmin: 1
NOTICE:  Materialized view acl_mv is stale. Refreshing...
NOTICE:  Materialized view acl_mv refreshed successfully.
 count 
-------
 14027
(1 row)

Time: 19528.798 ms (00:19.529)


omnigres=# select count(*) from acl_co;
NOTICE:  Checking materialized view: acl_mv, last_max_xid: 38056, current_max_xmin: 38056
NOTICE:  Materialized view acl_mv is not stale. No refresh needed.
 count 
-------
 14027
(1 row)

Time: 30.394 ms

omnigres=# select count(*) from acl_co;
NOTICE:  Checking materialized view: acl_mv, last_max_xid: 1, current_max_xmin: 1
NOTICE:  Materialized view acl_mv is not stale. No refresh needed.
 count 
-------
 14063
(1 row)

Time: 28.877 ms
omnigres=# 
```

yrashk:
First of all, thank you for the submission.

Can you please follow the guidelines with regards to the target branch (`next/omni_schema`), the workflow and commit message format?

As for the solution – we've tried something like this before – however, without the clever use of system columns (kudos for this – though this would definitely require documenting this use – its validity and safety)

The biggest problem with materializing the view that we saw is that it doesn't solve our core problem – we often have just one run, therefore there's no opportunity to use the cache to speed it up. If we run it just one time, caching is of no help.

As for changing what the queries return, I'd highly prefer if you document precisely what is changed and _why_. 

Observartion: `acl_co` refreshing takes quite a bit of time. What is the timing for the original method? (I have no point of reference here as all these queries are significantly faster on my hardware – under 500ms)

extensions/omni_schema/src/meta/catalog.sql
```diff
@@ -2093,298 +2093,429 @@ create view obj_object_id as
...
+    lock_acquired := pg_try_advisory_lock(hashtext(mv_name));  -- Try to acquire the lock
+
+    IF NOT lock_acquired THEN
+        RAISE NOTICE 'Could not acquire lock for materialized view %, skipping refresh.', mv_name;
+        RETURN;  -- Skip refresh if lock could not be acquired
+    END IF;
+
+    BEGIN
+        -- 5. Connect to the remote database using dblink
+        PERFORM dblink_connect(format('dbname=omnigres user=%s', current_user));
```
yrashk, line 2358: Hard-coding here won't work for the general case.

extensions/omni_schema/src/meta/catalog.sql
```diff
@@ -2093,298 +2093,429 @@ create view obj_object_id as
...
+
+            -- Refresh the materialized view concurrently
+            IF mv_name = 'acl_mv' THEN
+                PERFORM dblink_exec('REFRESH MATERIALIZED VIEW CONCURRENTLY omni_schema.acl_mv');
+            ELSIF mv_name = 'dependency_pre_mv' THEN
+                PERFORM dblink_exec('REFRESH MATERIALIZED VIEW CONCURRENTLY omni_schema.dependency_pre_mv');
+            END IF;
+
+            -- Update the refresh log with the new last_max_xid (current_max_xmin in this case)
+            UPDATE mv_refresh_log
```
yrashk, line 2372: How do we know here the matterialized view is refreshed at this point, since it is done concurrently?

Also, what are the implications of doing this refreshing in a separate backend?

extensions/omni_schema/src/meta/catalog.sql
```diff
@@ -2093,298 +2093,429 @@ create view obj_object_id as
...
+AS $function$
+DECLARE
+    current_max_xid bigint;
+    last_max_xid bigint;
+    current_max_xmin bigint;
+    is_refresh_in_progress boolean;
+    lock_acquired boolean;
+BEGIN
+    -- 1. Get the last stored max XID from the log table for the specified mv_name
+    SELECT mv_refresh_log.last_max_xid INTO last_max_xid
```
yrashk, line 2301: Below is a clever use of system columns. Do you think it'll work for the case of querying this view within the same transaction – in the presence of catalog changes?

extensions/omni_schema/src/meta/catalog.sql
benqethan: Good catch!  I will add a status polling to make sure the aync task is done. Alternatively, will try remove 'concurrently'(but i remember the view cannot refresh the mv while try to read from it without concurrently)

The reason to do it in a separate backend is the view cannot refresh the mv in a same active session(even do it concurrently). Before I use dblink, i had tried:
- db_background. It works, but not official contrib from the  postgres
- event trigger. That's the one make more sense and most straightforward. but it will slow down all DDL operation. Also I realized that we should avoid any changes to system catalog tables(such as index, trigger) etc

extensions/omni_schema/src/meta/catalog.sql
benqethan: will fix

extensions/omni_schema/src/meta/catalog.sql
benqethan: That's valid concern. we can add a snapshot flush at beginning. But that will slow down the mv refresh. Considering view query is not that heavy due to the nature of the metadata, maybe the accuracy is the way it should go.

benqethan:
@yrashk Thanks for the review. Will go thru the guideline and also check the failed tests.

Whether mv is effective or not, I believe it depends on the usage pattern of the views. I think for metadata there is much more read than write/change from the platform aspect. For example, some job can keep checking some metadata change and act on it if there is.  From that sense, mv's cache could help. Of course if the use cases are quite different, we can consider to only optimize the query itself on the system catalog. But there is much less to leveage since any system level change(such as index, trigger etc) will put a negative impact to the database engine.

Let me know your thoughts.

benqethan:
Problem: omni_schema.dependency and omni_schema.acl views are slow 

Solution: 
1. Use materialized view to create a pre-filtered data for view, caches the dependency results for better performance
2. Functions to detect the system catalog changes and determine a mv refresh is needed
3. Use dblink to connect to db and perform the refresh.

benqethan:
> As for changing what the queries return, I'd highly prefer if you document precisely what is changed and _why_.

There is left join issue on original query:
left join pg_namespace lns on ns.oid = lt.typnamespace  -- Wrong!
left join pg_namespace rns on ns.oid = rt.typnamespace  -- Wrong!

This will return NULL for most type namespace lookups

left join pg_namespace lns on lns.oid = lt.typnamespace  -- Correct
left join pg_namespace rns on rns.oid = rt.typnamespace  -- Correct

benqethan:
> Observartion: `acl_co` refreshing takes quite a bit of time. What is the timing for the original method? (I have no point of reference here as all these queries are significantly faster on my hardware – under 500ms)

I may need more time to troubleshooting why it happens. There will be some overhead caused by staleness check, lock, mv refresh etc. But 19 sec seems a bit longer than I expected.

yrashk:
Thank you for the update: will look into it soon. The join bug discovery is worth a separate PR!

PR #840, opened 2025-03-31
Add coverage for omni_httpd passing listener id to handlers.

diogob:
Problem: We cannot trust the behaviour of passing the listener id to an omni_httpd handler won't change or break over time. We want to rely on this parameter to implement a gateway so that one listener can receive requests on behalf of all the others.

Solution: Add a test that make this behaviour explicit.


yrashk:
Might make sense to amend the documentation as well.

PR #841, opened 2025-04-01
Problem: There are situations where we want to access distinct orbs through the same listener.

diogob:
Solution: Have a way to automatically proxy http requests from one orb
to another.


docker/initdb-slim/999-ready.sql.templ
```diff
@@ -1,3 +1,7 @@
-update omni_httpd.listeners set port = 8081;
+insert into omni_httpd.listeners (address, port) values ('0.0.0.0', '8081');
+insert into omni_httpd.listeners_handlers (listener_id, handler_id)
```
yrashk, line 2: `listeners_handlers` is deprecated

PR #842, opened 2025-04-03
Problem: Using a url pattern containing a port will never match.

diogob:
We want to use ports in routes so we are able to select handlers based on listeners.
The same goes for scheme and host.

Solution: Include the port, scheme and host to the URL before matching it.


diogob:
Moved to #841

PR #880, opened 2025-06-22
docs: use correct port in quick start

tomberek:


yrashk:
Thank you for spotting this! 8081 is meant as a service port – this is where the `omnigres` database is exposed by default. 8080 is the convention to be used by other databases.

The idea here is that the `omnigres` database exposes a "service application" (think Postgres cluster API, dashboard, etc.) whereas other databases are other "applications".

(There's some intersection with this bit of documentation for the CLI: https://github.com/omnigres/cli/blob/master/docs/modules/ROOT/pages/index.adoc – please keep in mind that CLI is WIP and is due to some rewriting)

Happy to merge this as is, but if you want to work the above in (or need any clarifications), that'd be great too – let me know!

PR #892, opened 2025-07-31
First impression

henrikingo:


README.md
```diff
@@ -33,6 +33,14 @@ Omnigres makes Postgres a developer-first application platform. You can deploy a
 The fastest way to try Omnigres out is by using its [container image](https://github.com/omnigres/omnigres/pkgs/container/omnigres):
 
 ```shell
+# First time only:
```
yrashk, line 36: Perhaps it would rather make more sense to point to an article (or articles) on how to install these – if none are good, a separate markdown file in our repo, then?

docker/entrypoint.sh
```diff
@@ -12,4 +12,15 @@ if [ -n "$TAILSCALE_AUTHKEY" ]; then
...
\ No newline at end of file
+
+PGUSER="${PGUSER:-$POSTGRES_USER}"
+echo "Now starting Omnigres database and application services. (${POSTGRES_INITDB_ARGS})"
+echo "You can connect to this service the following ways:"
+echo
+echo "    psql -h localhost -p ${PGPORT:-5432} -U ${PGUSER:-postgres} omnigres"
+echo "    http://localhost:8081/"
+echo
+echo "If you're using docker, verify the above ports with 'docker ps'"
```
yrashk, line 23: I like this idea! I wonder if we can retrieve the port mappings from within the container – or not worth it?

Arguably, this always applies:
> echo "If you're using docker, verify the above ports with 'docker ps'"

Because it's part of the docker image entrypoint.

PR #904, opened 2025-08-05
Problem: lack of useful data (🚀 omni_datasets 0.1.0)

henrikingo:
This makes it more difficult to test problems with realistic or semi-realistic data as it has to be procured.

Solution: start omni_datasets extension

We'll start with this:

This extension installs the legendary Northwind database, that was used as a demo database in Microsoft Access and later also in SQL Server.

The idea is to use it as a general test dataset that other extensions can depend on. This extension can also bring more data with more modern features. For example, some vector data or useful open data – through instantiation technique.


henrikingo:
Ok @yrashk , here it is. Thanks for clarifying guidance on how to create extensions. I then also ditched the use of docker and just ran the tests from build/ directory as explained in the end of the README.

extensions/omni_datasets/mkdocs.yml
```diff
@@ -0,0 +1,2 @@
+INHERIT: ../../mkdocs.base.yml
+site_name: omni_northwind
```
henrikingo, line 2: omni_dataset

yrashk: fixed this one

extensions/omni_datasets/tests/northwind.yml
```diff
@@ -0,0 +1,70 @@
...
+instance:
+  config:
+    shared_preload_libraries: */env/OMNI_SO
+    max_worker_processes: 64
+    omni_worker.workers: 10
+  init:
+  - create schema if not exists omni_datasets
+  - create extension omni_datasets cascade
+  - create schema northwind
+  - select omni_datasets.instantiate_northwind('northwind')
```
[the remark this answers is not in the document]
yrashk: It's not possible to use `regnamespace` before the schema is created. One can use `name` but then we'd have to check if the schema does exist, and if it doesn't do `execute format('create schema %I', schema);` and so on.

Both approaches are fine, and I am not particularly settled into any – they both have pros and cons.

PR #925, opened 2025-08-26
Problem: omni_kube always lists all resources

yrashk:
Solution: add field and label selectors


yrashk:
Missing:

- [ ] docs
- [ ] tests

PR #928, opened 2025-08-27
Problem: omni_kube API doesn't compose well

yrashk:
For example, api() special-cases things like fieldSelector and labelSelector and it's going to be hard to compose watches.

Solution: extract path helpers and make them composable


yrashk:
Missing:

- [ ] docs
- [ ] extra tests

PR #934, opened 2025-08-31
Problem: omni's shared memory management (🚀 omni_shmem 0.1.0, apex 0.1.0)

yrashk:
It's based on DSA/DSM and is, frankly, tedious to follow. It's also coupled with omni's release lifecycle.

Solution: reimplement it as a new extension

Simpler design, based on Boost Interprocess (wrapped into our lightweight Oink library).

Importantly, this introduces APEX: API for Postgres Extensions. It's a way to expose C APIs between extensions and a robust initialization technique. I plan to test it out with our extensions and then release it more publicly.

Also, on testing side, since we're doing this APEX business, I am also now testing C/C++ level code with doctest-based tests compiled into the extension. The idea here is that we'll probably disable those for full production builds.


yrashk:
This has been sitting here on the back burner a little too long. What has been delaying here is just hashing of the API. While important, it's not 100% critical to the release of the first version.

Let's acknolwedge this deficiency and release it.

PR #956, opened 2025-09-16
Problem: retaining old records

yrashk:
Our users have needs to retain records, both for compliance and oeprational needs.

Solution: kickstart omni_audit extension


yrashk:
This has been marinating. Let's release this as a preview. It's not perfect but it does something.

PR #965, opened 2025-10-25
Problem: omni_schema.dependency and omni_schema.acl views are slow

zelosleone:
- First, we had a cartesian product in the dependency view for `omni_schema.dependency`. We were joining operator namespace (`ns`) with type namespace instead of joining type namespace table (`lns`) with type's namespace field. The correct, sound  result should have been 32 rows. But it was doing 32 x 22 x 22 resulting in extra, completely not needed and not sound 15,488 rows which resulted in 94.9 more ms!
- We were doing materialization early but was also doing zero filtering for the table. Postgres usually handles this if it can but here, it couldn't. Added early filtering with `deptype != 'i'`
- Eliminated the need for full composite type, this also boosted our speed since we didn't need the full type. We only needed `select id`, `d.refclassid`, `d.refobjid`, `d.refobjsubid` fields.
- For acl, it didn't have low hanging fruit like these. But introducing pre-computation to it also made a x2.7 speed boost  because the actual speed blockage was from calling the entire acl view for every function, every time and trying to compute from the full view for every function. Now we compute once, then re-use it for every function that needs it.
- `omni_schema.dependency` also got a x2.1 speed boost. From around 266 ms to 128 ms.
- Using `BUFFER` and `EXPLAIN` together is a bliss.

/claim #823

Would love to hear your feedback on this @yrashk


zelosleone:
@yrashk Any updates on this?

yrashk:
Thank you for this PR! The operator namespace bug fix is a genuine correctness improvement - the original code was incorrectly joining `ns.oid` with `lt.typnamespace` and `rt.typnamespace` instead of using `lns.oid` and `rns.oid`, which would cause incorrect results when operators and their operand types are in different schemas. This fix is valuable regardless of performance considerations.

Could you share more details about how you measured the performance improvements? Specifically:
- What was your test environment (Postgres version, hardware)?
- What data/schema did you use for benchmarking?
- How many runs did you perform to get consistent measurements?

---

## Performance Testing Results

I ran performance tests on both the `master` branch and this PR branch to verify the claimed improvements.

### Methodology

1. Created a heavy test scenario with:
   - 3 test schemas
   - 150 custom types (50 per schema)
   - 300 functions (100 per schema)
   - 90 tables with serial columns (30 per schema)
   - Cross-schema operators
   - 20 views depending on tables

2. Ran each view query 5 times after a warm-up run to get consistent measurements

3. Environment: PostgreSQL 18.0, macOS (Darwin), Apple Silicon

### Results

#### `omni_schema.dependency` View

| Run | Master | PR Branch | Ratio |
|-----|--------|-----------|-------|
| 1 | 972ms | 547ms | 1.78x |
| 2 | 926ms | 555ms | 1.67x |
| 3 | 938ms | 552ms | 1.70x |
| 4 | 926ms | 550ms | 1.68x |
| 5 | 957ms | 557ms | 1.72x |
| **EXPLAIN ANALYZE** | **980ms** | **573ms** | **1.71x** |

**Average improvement: ~1.71x faster** ✅

#### `omni_schema.acl` View

| Run | Master | PR Branch | Ratio |
|-----|--------|-----------|-------|
| 1 | 229ms | 387ms | 0.59x |
| 2 | 252ms | 388ms | 0.65x |
| 3 | 260ms | 390ms | 0.67x |
| 4 | 260ms | 385ms | 0.68x |
| 5 | 289ms | 385ms | 0.75x |
| **EXPLAIN ANALYZE** | **239ms** | **386ms** | **0.62x** |

**Average: ~1.5x slower (regression)** ❌

### Summary

- The `dependency` view shows a **~1.7x improvement**, which is good but below the bounty's 2x minimum requirement
- The `acl` view shows a **~1.5x regression** - the CTE pre-computation (`callable_sigs`) appears to add overhead rather than reduce it in this test scenario
- The correctness fix for the operator namespace joins is valid and should be kept

Would you be able to share your test methodology so we can understand the discrepancy in results? It's possible the ACL optimization performs differently depending on the data characteristics.

---

<details>
<summary>Performance Test Script</summary>

```sql
-- Heavy performance test for omni_schema views
-- Creates many objects to stress the dependency and acl views

-- First create the extension
create extension omni_schema cascade;

-- Create multiple schemas
create schema test_schema1;
create schema test_schema2;
create schema test_schema3;

-- Create many types across schemas
do $$
begin
    for i in 1..50 loop
        execute format('create type test_schema1.test_type_%s as (id int, name text, value numeric)', i);
        execute format('create type test_schema2.test_type_%s as (id int, name text, value numeric)', i);
        execute format('create type test_schema3.test_type_%s as (id int, name text, value numeric)', i);
    end loop;
end $$;

-- Create many functions
do $$
begin
    for i in 1..100 loop
        execute format('create function test_schema1.test_func_%s(a int, b text) returns int language sql as $f$ select a $f$', i);
        execute format('create function test_schema2.test_func_%s(a int, b text, c numeric) returns text language sql as $f$ select b $f$', i);
        execute format('create function test_schema3.test_func_%s(a int) returns int language sql as $f$ select a * 2 $f$', i);
    end loop;
end $$;

-- Create tables with various columns
do $$
begin
    for i in 1..30 loop
        execute format('create table test_schema1.test_table_%s (id serial primary key, name text, created_at timestamp default now(), data jsonb)', i);
        execute format('create table test_schema2.test_table_%s (id serial primary key, description text, amount numeric(10,2))', i);
        execute format('create table test_schema3.test_table_%s (id serial primary key, ref_id int, status text)', i);
    end loop;
end $$;

-- Create some operators with cross-schema types
create function test_schema1.type_eq(test_schema1.test_type_1, test_schema2.test_type_1) returns boolean language sql as $$ select true $$;
create operator test_schema3.=== (leftarg = test_schema1.test_type_1, rightarg = test_schema2.test_type_1, procedure = test_schema1.type_eq);

create function test_schema2.type_lt(test_schema1.test_type_2, test_schema3.test_type_2) returns boolean language sql as $$ select true $$;
create operator test_schema1.<<< (leftarg = test_schema1.test_type_2, rightarg = test_schema3.test_type_2, procedure = test_schema2.type_lt);

-- Create views that depend on tables
do $$
begin
    for i in 1..20 loop
        execute format('create view test_schema1.test_view_%s as select * from test_schema1.test_table_%s', i, (i % 30) + 1);
    end loop;
end $$;

-- Warm up
select count(*) from omni_schema.dependency;
select count(*) from omni_schema.acl;

\timing on

-- Run dependency tests (5 times each)
select 'dependency run 1' as test;
select count(*) from omni_schema.dependency;

select 'dependency run 2' as test;
select count(*) from omni_schema.dependency;

select 'dependency run 3' as test;
select count(*) from omni_schema.dependency;

select 'dependency run 4' as test;
select count(*) from omni_schema.dependency;

select 'dependency run 5' as test;
select count(*) from omni_schema.dependency;

-- Run acl tests (5 times each)
select 'acl run 1' as test;
select count(*) from omni_schema.acl;

select 'acl run 2' as test;
select count(*) from omni_schema.acl;

select 'acl run 3' as test;
select count(*) from omni_schema.acl;

select 'acl run 4' as test;
select count(*) from omni_schema.acl;

select 'acl run 5' as test;
select count(*) from omni_schema.acl;

-- Explain analyze for detailed timing
select 'EXPLAIN dependency' as test;
explain (analyze, buffers, timing) select count(*) from omni_schema.dependency;

select 'EXPLAIN acl' as test;
explain (analyze, buffers, timing) select count(*) from omni_schema.acl;
```

</details>

PR #973, opened 2026-02-10
Problem: omni_txn.retry max_attempts is not always practical

### LINKED ISSUE #669: Problem: omn_txn.retry max_attempts is not always practical

buildingvibes:
Fixes #669

## Problem

`omni_txn.retry` only supports `max_attempts` to control how many times a serialized transaction will be retried. This doesn't tell us how long we should be trying to commit the transaction for. In time-sensitive scenarios, expressing a deadline as a duration is more practical than a fixed retry count.

## Solution

Introduce an optional `timeout` interval parameter to `omni_txn.retry`. When provided, the retry loop checks elapsed wall-clock time after each serialization failure and stops retrying once the timeout has been exceeded, raising a `serialization_failure` error with the message `retry timeout has been reached`.

The parameter defaults to `NULL` (disabled), preserving full backward compatibility. Both `timeout` and `max_attempts` can be used together — whichever limit is reached first will stop the retries.

### Changes

- **`extensions/omni_txn/retry.c`**: Added timeout parsing (argument index 6), `interval_to_microsecs` conversion helper, wall-clock tracking via `GetCurrentTimestamp()`, and timeout check in the retry loop.
- **`extensions/omni_txn/migrate/7_retry_timeout.sql`**: New migration that drops and recreates the `retry` procedure with the additional `timeout interval default null` parameter.
- **`extensions/omni_txn/tests/retry.yml`**: Two new tests — one verifying timeout triggers correctly with a 10ms timeout, and one verifying `NULL` timeout doesn't interfere.
- **`extensions/omni_txn/CHANGELOG.md`**: Added 0.6.0 entry.
- **`versions.txt`**: Bumped `omni_txn` from 0.5.0 to 0.6.0.

### Usage

```sql
-- Retry for at most 5 seconds
call omni_txn.retry('UPDATE accounts SET balance = balance - 100 WHERE id = 1', timeout => interval '5 seconds');

-- Combine with max_attempts
call omni_txn.retry('UPDATE accounts SET balance = balance - 100 WHERE id = 1', max_attempts => 50, timeout => interval '2 seconds');
```

### Alternatives considered

- A GUC variable for timeout — rejected in favor of per-call control, consistent with how `max_attempts` works.
- A separate `retry_with_timeout` function — unnecessary complexity when a single optional parameter suffices.


buildingvibes:
Closing this PR as the project appears to be less active recently. Happy to reopen if there's interest in reviewing this. Thank you\!

